PR #8, opened 2017-09-07
Add sentry support

schnapster:
Add support for sentry.io


freyacodes:
What happens if no key is given?

schnapster:
Nothing.

freyacodes (approved):

PR #17, opened 2017-10-21
Add custom penalty system

Fabricio20:
Adds a custom penalty system on the load balancer.


LavalinkClient/src/main/java/lavalink/client/io/LavalinkLoadBalancer.java
```diff
@@ -22,22 +22,26 @@
...
     private Lavalink lavalink;
     //private Map<String, Optional<LavalinkSocket>> socketMap = new ConcurrentHashMap<>();
+    private List<BalancerPenalty> customPenalties = new ArrayList<>();
 
     LavalinkLoadBalancer(Lavalink lavalink) {
         this.lavalink = lavalink;
     }
 
-    public LavalinkSocket determineBestSocket() {
+    public LavalinkSocket determineBestSocket(Long guild) {
```
freyacodes: We should preferably use unboxed primitives

LavalinkClient/src/main/java/lavalink/client/io/LavalinkLoadBalancer.java
```diff
@@ -52,6 +56,14 @@ public LavalinkSocket determineBestSocket() {
         return leastPenalty;
     }
 
+    public void addPenalty(BalancerPenalty penalty) {
+        this.customPenalties.add(penalty);
+    }
+
+    public void remPenalty(BalancerPenalty penalty) {
```
freyacodes: Should be remove instead of rem for consistency and brevity

LavalinkClient/src/main/java/lavalink/client/io/LavalinkLoadBalancer.java
```diff
@@ -138,4 +167,11 @@ public String toString() {
         }
     }
 
+    public interface BalancerPenalty {
```
freyacodes: This is a really misleading name. Maybe name it something like `PenaltyProvider`.

Would also be nice if you moved it to its own file.

LavalinkClient/src/main/java/lavalink/client/io/LavalinkLoadBalancer.java
```diff
@@ -66,23 +78,25 @@ void onNodeConnect(LavalinkSocket connected) {
...
     @SuppressWarnings("unused")
     public static class Penalties {
 
         private LavalinkSocket socket;
+        private final Long guild;
         private int playerPenalty = 0;
         private int cpuPenalty = 0;
         private int deficitFramePenalty = 0;
         private int nullFramePenalty = 0;
+        private List<BalancerPenalty> custom;
```
freyacodes: This should be computed as a value on construction, rather than being stored and invoked later

freyacodes (changes requested):
Optional: If you want to make this into a really good PR, you could migrate all the existing penalty values (cpu, players, etc) to the "custom" system.

LavalinkClient/src/main/java/lavalink/client/io/PenaltyProvider.java
```diff
@@ -0,0 +1,12 @@
+package lavalink.client.io;
+
+/**
+ * Copyright (C) Fabricio20 2017 - All Rights Reserved.
```
freyacodes: 😅

Fabricio20:
I don't think moving the default penalties to PenaltyProvider classes will be helpful, considering this way we keep less objects in-memory and we can still fetch the value of each default penalty for use on custom ones.

LavalinkClient/src/main/java/lavalink/client/io/PenaltyProvider.java
```diff
@@ -0,0 +1,12 @@
...
+ * Created by Fabricio20 on 2017-10-21.
+ */
+public interface PenaltyProvider {
+
+    // Using the penalties class allows for fetching default ones like CPU or Players.
```
freyacodes: This should be javadoc

LavalinkClient/src/main/java/lavalink/client/io/LavalinkLoadBalancer.java
```diff
@@ -30,13 +30,13 @@
 
     private Lavalink lavalink;
     //private Map<String, Optional<LavalinkSocket>> socketMap = new ConcurrentHashMap<>();
-    private List<BalancerPenalty> customPenalties = new ArrayList<>();
+    private List<PenaltyProvider> customPenalties = new ArrayList<>();
```
freyacodes: Nitpick, but that field name could be better

freyacodes (changes requested):

freyacodes:
>I don't think moving the default penalties to PenaltyProvider classes will be helpful, considering this way we keep less objects in-memory and we can still fetch the value of each default penalty for use on custom ones.

That's irrelevant, we don't even store these penalties. This is grains of sand compared to the ocean of resources used to play audio.

LavalinkClient/src/main/java/lavalink/client/io/PenaltyProvider.java
```diff
@@ -1,12 +1,35 @@
...
- * Copyright (C) Fabricio20 2017 - All Rights Reserved.
- * Created by Fabricio20 on 2017-10-21.
- */
 public interface PenaltyProvider {
 
-    // Using the penalties class allows for fetching default ones like CPU or Players.
+    /**
+     * Using the penalties class allows for fetching default penalties like CPU or Players.
+     *
+     * @param balancer - Instance of Penalties class representing the node to check.
```
freyacodes: This is more nitpicking, but I feel like this javadoc doesn't describe the purpose of this interface and method. Balancer is a misleading name, as the Penalties class is not a load balancer in itself. Using `{@Link ... }` would be nice

freyacodes:
I also notice now that this is based on the wrong branch. I'll attempt to change the base

freyacodes (approved):

PR #19, opened 2017-10-25
Allow setting of a custom buffer

schnapster:
Idk how to handle Spring properties properly but this doesn't throw any exceptions when anything is missing in the config.

This is an "experimental" feature based on this PR https://github.com/sedmelluq/jda-nas/pull/5 which I pulled into our jda-nas fork already. We talked about wanting this months ago, but never actually implemented it, so here it is.


freyacodes (approved):
lgtm

PR #21, opened 2017-10-26
Add async packet provider

Shredder121:


LavalinkServer/src/main/java/lavalink/server/Launcher.java
```diff
@@ -130,7 +132,9 @@ public static void main(String[] args) {
                 log.info("Using default buffer");
                 nativeAudioSendFactory = new NativeAudioSendFactory();
             }
-            AudioConnection.setAudioSendFactory(nativeAudioSendFactory);
+            AudioConnection.setAudioSendFactory(
+                    AsyncPacketProviderFactory.adapt(nativeAudioSendFactory)
+            );
             log.info("JDA-NAS supported system detected. Enabled native audio sending.");
         } else {
             log.warn("This system and architecture appears to not support native audio sending! "
```
Shredder121, line 140: We could also wrap the original audio send factory in the async wrapper?
So the normal audio sending logic is also non blocking.

freyacodes: I'm not sure how that works

schnapster (approved):
lgtm

freyacodes:
Still doesn't look like sending is fast enough beyond ~560 players

PR #24, opened 2017-10-28
Added remote debugging

freyacodes:
This allows the execution of JavaScript for the sake of debugging


schnapster (approved):
As long as this is used via SSH tunnel and the firewall is up for the REST port this is fine.

Frederikam:
I can make this only work over localhost

PR #26, opened 2017-11-02
Added Feedback Incase Resolve Fails

ThePrimedTNT:
Added an event TrackResolveErrorEvent, this will be called if LavaLink fails to resolve a track


freyacodes:
This should be documented in IMPLEMENTATION.md

PR #29, opened 2017-12-01
Bump version on japp

Shredder121:


Shredder121:
This changed the clock leap logs to debug and actually named the threads it creates.
One bigger change is that the pool it used (which wasn't sized at all, because no good timeout period for tearing down the threads) is now removed and it will use threads as it needs them, the threads tearing themselves down as they finish.

freyacodes (approved):

PR #35, opened 2017-12-06
Fix setPause bug (and fixed not reconnecting after changing node)

Repulser:


LavalinkClient/src/main/java/lavalink/client/io/Link.java
```diff
@@ -251,6 +251,7 @@ public void destroy() {
     private void changeNode0(LavalinkSocket newNode) {
         currentNode = newNode;
         pendingNode = null;
+        connect(getCurrentChannel());
```
freyacodes, line 254: Has this been tested?

Repulser: Indeed, works perfectly

freyacodes (approved):

PR #43, opened 2017-12-12
Socket null fix

Repulser:


MrJacz:
Why not just make one pr for all changes you're going to make?

freyacodes:
>Why not just make one pr for all changes you're going to make?

PRs should change one thing at a time to make them easy to merge and review.

MrJacz:
I guess you guys have a different way of handling things then :)

schnapster:
I was summoned by a proposed removal of a `@Nonnull` annotation without a replacement :eyes:

Repulser:
@napstr Well the issue is it could return null if we choose not to select one if absent, whats the benefit in that annotation anyways?

freyacodes (approved):

freyacodes:
@Repulser Some IDEs will warn you of potential NPEs when you don't already check

PR #44, opened 2017-12-13
Fix shard disconnecting events

Repulser:
Tested


freyacodes:
Why is this required? IIRC this would be automatically handled by the server

Repulser:
Similar to guild leave, JDA Audio has no way of knowing and gets stuck, ive tested this

freyacodes:
You're right, we spoke about this. I must still have been sleeping

freyacodes (approved):

PR #48, opened 2017-12-21
Fix position in streams and use isSeekable

Repulser:


freyacodes (approved):

LavalinkClient/src/main/java/lavalink/client/player/LavalinkPlayer.java
```diff
@@ -135,7 +135,6 @@ public boolean isPaused() {
     @Override
     public long getTrackPosition() {
         if (getPlayingTrack() == null) throw new IllegalStateException("Not currently playing anything");
-        if (getPlayingTrack().getInfo().isStream) return Long.MAX_VALUE;
```
schnapster, line 138: Relying on this return value might be a thing that is or was hard coded in FredBoat :^)
Probably should be checked before updating lavalink the next time. I might if I don't forget to later.

PR #54, opened 2018-01-16
Server side magic :)

Repulser:


LavalinkServer/src/main/java/lavalink/server/io/ConnectionManagerImpl.java
```diff
@@ -0,0 +1,39 @@
...
+        this.socket = socket;
+    }
+
+    @Override
+    public void removeAudioConnection(String guildId) {
+    }
+
+    @Override
+    public void queueAudioConnect(String guildId, String channelId) {
+        log.warn("queueAudioConnect was requested, this shouldn't happen");
```
freyacodes: Log the method parameters so that this can be traces

LavalinkServer/src/main/java/lavalink/server/io/CoreClientImpl.java
```diff
@@ -22,171 +22,41 @@
...
 
     @Override
     public void sendWS(String message) {
-        log.info(message);
-        JSONObject json = new JSONObject();
-        json.put("op", "sendWS");
-        json.put("shardId", shardId);
-        json.put("message", message);
-        socket.send(json.toString());
+        log.warn("sendWS was requested, this shouldn't happen");
```
freyacodes: Same applies to this class

freyacodes (changes requested):

freyacodes (approved):

LavalinkServer/src/main/java/lavalink/server/io/CoreClientImpl.java
freyacodes: You forgot this still

freyacodes (approved):

PR #58, opened 2018-01-22
Add a client side Prometheus metrics collector

schnapster:
Also added a name property for nodes

### DIFF
--- a/LavalinkClient/src/main/java/lavalink/client/io/metrics/LavalinkCollector.java
+++ b/LavalinkClient/src/main/java/lavalink/client/io/metrics/LavalinkCollector.java
@@ -1,0 +1,105 @@
+package lavalink.client.io.metrics;
+
+import io.prometheus.client.Collector;
+import io.prometheus.client.GaugeMetricFamily;
+import lavalink.client.io.Lavalink;
+import lavalink.client.io.LavalinkSocket;
+import lavalink.client.io.RemoteStats;
+
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * Created by napster on 22.01.18.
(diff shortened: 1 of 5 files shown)


LavalinkClient/src/main/java/lavalink/client/io/metrics/LavalinkCollector.java
```diff
@@ -0,0 +1,105 @@
...
+            uptimeSeconds.addMetric(labels, stats.getUptime() / 1000);
+
+            memFree.addMetric(labels, stats.getMemFree());
+            memUsed.addMetric(labels, stats.getMemUsed());
+            memAllocated.addMetric(labels, stats.getMemAllocated());
+            memReservable.addMetric(labels, stats.getMemReservable());
+
+            cpuCores.addMetric(labels, stats.getCpuCores());
+            systemLoad.addMetric(labels, stats.getSystemLoad());
+            lavalinkLoad.addMetric(labels, stats.getLavalinkLoad());
```
freyacodes, line 96: I'm wondering if I should remove this, as it has always been broken

schnapster: If you do, you can just delete the code here.
I think however it would be better to fix the stats. Should be easy enough, because we can just ask the JVM directly for them, instead of going through oshi.

freyacodes:
Where does the JVM provide those stats?

schnapster:
Rip, it doesn't. We have the stats in our dashboard, cause prometheus does some hax.
See here: https://github.com/prometheus/client_java/blob/master/simpleclient_hotspot/src/main/java/io/prometheus/client/hotspot/StandardExports.java#L64

freyacodes (approved):

Issue #41, opened 2017-12-10
[Suggestion] [REST] Way to change YouTube Playlist size limit


devoxin:
So it would appear that the current limit is lavaplayer's default (600) tracks. From what I've seen, there's no property to change this.

Perhaps a property that can be changed in `application.yml` to set the maximum amount of tracks loaded from a playlist?

freyacodes:
Wouldn't be difficult to accomplish with the config file. I believe this limit only applies to the youtube source provider

PR #60, opened 2018-02-01
Add ability to configure playlist import limit

devoxin:
This PR will allow users to change how many videos are imported from a YouTube playlist via `application.yml`

Fulfills #41


freyacodes (changes requested):
Looks good but there should be an example in the application.yml file. Also I'm redirecting this to the dev branch.

freyacodes (approved):

PR #63, opened 2018-02-06
Fix redundancy, update deps and rely on local player count

Repulser:


LavalinkClient/src/main/java/lavalink/client/io/Link.java
```diff
@@ -149,7 +149,7 @@ public void changeNode(LavalinkSocket newNode) {
     void onDisconnected() {
         setState(State.NOT_CONNECTED);
         LavalinkSocket socket = getNode(false);
-        if (socket != null) {
+        if (socket != null && state != State.DESTROYING && state != State.DESTROYED) {
```
freyacodes, line 152: What is the reason for this new state? It seems redundant

LavalinkClient/src/main/java/lavalink/client/io/LavalinkLoadBalancer.java
```diff
@@ -122,6 +128,11 @@ private Penalties(LavalinkSocket socket, long guild, List<PenaltyProvider> penal
             penaltyProviders.forEach(pp -> customPenalties += pp.getPenalty(this));
         }
 
+        private int getLocalPlayingPlayers() {
```
freyacodes: Local players? These players are remote

freyacodes (changes requested):

LavalinkClient/src/main/java/lavalink/client/io/LavalinkLoadBalancer.java
```diff
@@ -122,6 +128,11 @@ private Penalties(LavalinkSocket socket, long guild, List<PenaltyProvider> penal
...
+            Long players = lavalink.getLinks().stream().filter(link -> socket.equals(link.getNode(false)) && link.getPlayer().getPlayingTrack() != null && !link.getPlayer().isPaused()).count();
```
freyacodes: This looks like it was meant to return the playing players of a specific node, not all. Also please avoid long one-liners like this. Use linebreaks.

freyacodes (approved):

PR #74, opened 2018-02-17
Add LavalinkServer Dockerfile

itslukej:
This is to be built assuming you have compiled the jars with gradle initally, as discussed with Napster this could be used for an official docker image in the future.

`docker build -f LavalinkServer/docker/Dockerfile -t lavalink .`


freyacodes:
I believe it is bad practice to hardcode the config values into the image

itslukej:
I have put the defaults into the dockerfile, they are meant to be overidden using the `-e` flag on `docker run` e.g `docker run -e LAVALINK_SERVER_PASSWORD="youshallpass" lavalink`. My mistake for not mentioning that

freyacodes:
Better practice would be to use volumes

schnapster:
The goal here was to use OS environment vars as the main source of the configuration.
Seeing that OS environment vars override any application.yaml, yeah, packaging the application.yml.example as application.yml and exposing it as a volume could be an option to bring both worlds together.

iCrawl:
I disagree that volumes are a better practices. Docker containers should not have an open link to the host if its not needed. And considering Lavalink is pretty stateless in itself it would be a waste.

freyacodes (approved):

PR #78, opened 2018-02-24
Update README.md

IceeMC:


freyacodes:
This line is meant to give a sense of scale of how widely used Lavalink is. These bots all have over 100,000 guilds, so it doesn't make a lot of sense to add much smaller bots.

Issue #38, opened 2017-12-08
[Suggestion] [REST] Search types in loadtracks


devoxin:
Perhaps a few fields (such as the ones below) could be included in the response?

| Field | Type | Description |
| ----- | ----- | ------------- |
| isSearchResult | boolean | Whether or not the results are from a search |
| provider | string | Which provider returned the tracks (e.g. YouTube, Vimeo etc) |
| isPlaylist | boolean | Whether or not the results are from a playlist |

PR #80, opened 2018-02-27
Add isPlaylist to response JSON

devoxin:
This allows clients to identify URL or Playlist results and handle the response accordingly.

**This is a breaking change to clients as it modifies the JSON that is returned**

Partially fulfills #38


freyacodes:
This could use some documentation for implementation

freyacodes (approved):
Looks good to me, but I don't believe we are ready to make a new release at the moment. TeamCity has built the artefact for this PR, so you will be able to use that in the meanwhile.

freyacodes:
Closing this as a merge is not possible when the downstream repo is deleted.

PR #82, opened 2018-03-02
Put all the dependency versioning into the parent file

schnapster:


schnapster:
Updated to account for #81

freyacodes (approved):

PR #85, opened 2018-03-05
Bump dependencies and gradle

schnapster:
Bump all dependencies and gradle


schnapster:
I noticed a suboptimal usage of gradle tasks in the project.
After this is merged, it is required to add an additional task to the CI server (gradle install **bootJar**) so it keeps producing server jar artifacts.
I will do another PR combining everything that needs to be done under a `gradle build` task. The reason I'm not piling it into this PR is because it would also require fixing tests of the client which currently arent run anywhere in the pipeline (lolwat, yeah)

schnapster:
With #87 merged I changed the CI gradle task to `gradle build`.
The v1-legacy branch builds, the dev branch builds too.
Master is broken. It was technically broken all the time, its just that the CI wasn't attempting to compile the broken tests. Since master is only used to update docs / README, I guess that's an acceptable state until the next release?

schnapster:
Added a commit that makes sure (client) tests are not only compiled, but also actually executed.
In the process rewrote parts of the voicechannel connection test to be less v1 and more stable.
However that whole test class relies on a running LavalinkServer (so its actually an integration test, not a unit test), so it won't be executed on the CI (or even locally) unless it finds the required system properties to be set.

freyacodes (approved):

PR #92, opened 2018-04-13
Volume fix

Repulser:
Fixed a bug where volume doesnt carry over if set before playing and bumped dependencies


freyacodes:
Fixes https://github.com/Frederikam/FredBoat/issues/458

PR #93, opened 2018-04-14
Fix exception thrown due to websocket port being in use when restarting

schnapster:
Reported by @brussell98:
image

Reproduced by SIGKILLING a lavalink that had a client connected.

Could not reproduce in several tries after adding the that is attached to this PR.

Related SO: https://stackoverflow.com/questions/31864369/java-net-bindexception-address-already-in-use-jvm-bind


shikhir-arora:
I've had this error before for the same reasons, usually if I restart Lavalink too quickly after a SIGKILL was sent.

freyacodes:
The port seems to be properly released after a few seconds. A solution could be to retry binding to it a few seconds apart.

freyacodes:
I'm an idiot. When I saw this in my mail I thought it was an issue, not a PR.

PR #100, opened 2018-05-07
Lavalink.py also supports V3 :)

devoxin:


AnotherGenZ:
I believe Fred doesn't like having the same client listed multiple times, even if it's for different versions. This was the case for SandySounds at least.

freyacodes:
v3 isn't final yet

PR #102, opened 2018-05-18
Fix VSU spam on node connect

schnapster:


LavalinkClient/src/main/java/lavalink/client/io/LavalinkLoadBalancer.java
```diff
@@ -75,6 +75,13 @@ void onNodeDisconnect(LavalinkSocket disconnected) {
     }
 
     void onNodeConnect(LavalinkSocket connected) {
+        long otherAvailableNodes = lavalink.getNodes().stream()
```
freyacodes, line 78: Are you sure getNodes() only returns other nodes? Maybe filter out the new node just to be on the safe side.

schnapster: literally the next line...

freyacodes: Sorry, I thought that was another check to see if the node was connected

PR #103, opened 2018-05-20
[v3] Add custom sentry tags

schnapster:
By default sentry tags the server name. This PR allows to set optional additional custom tags for a lavalink server instance, which then can be sorted by in sentry. This PR is useful when running more than one LL instance on one machine for example, or otherwise group and sort through reports of various instances.

Moved the sentry configuration into its own configuration key / class to avoid cluttering the server configuration. The previous dsn location is deprecated, but still supported and will result in a friendly warning detailing the new location / configuration key, until it is removed in a version post v3.

### DIFF
--- a/LavalinkServer/application.yml.example
+++ b/LavalinkServer/application.yml.example
@@ -16,9 +16,14 @@ lavalink:
       mixer: true
       http: true
       local: false
-    sentryDsn: ""
     bufferDurationMs: 400
     youtubePlaylistLoadLimit: 600
+
+sentry:
+  dsn: ""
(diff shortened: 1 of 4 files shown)


schnapster:
Thought this one was not a breaking change either, but it kinda builds upon 25572625851b8b55603ad8eacd77c1356a5fa33a so...nope, not backporting it to dev.

freyacodes (approved):

freyacodes:
It kinda is a breaking change, as it changes the location of the DSN and deprecates the old one.

PR #104, opened 2018-05-20
[dev] Replace Guava/jsr305 with spotbugs nullability annotations

schnapster:
This PR removes the (rather big) Guava dependency. The only feature of that lib that we use are the jsr305 nullability annotations, which are a bad idea to use anyways due to the pollution of the `javax.annotations` namespace, and are bound to be removed from guava. These annotations have been replaced with spotbug's nullability annotations (which work exactly the same way).

See also:
https://github.com/google/guava/issues/2960
https://github.com/spotbugs/spotbugs/issues/130


freyacodes (approved):
LGTM

PR #105, opened 2018-05-20
[dev] Add prometheus metrics

schnapster:
Add basic metrics for logs and the JVM.

The metrics beans are fully conditional, so not opting in to metrics should have no impact™

No security has been added. Security is achieved by not exposing the host to public networks in the first place, or using a reverse proxy (and any security measures of the user's choice and the chosen reverse proxies' capability).


freyacodes:
This doesn't look like a breaking change. Would it make sense to include it in v2?

schnapster:
Yes. One moment.

schnapster:
:ok_hand:

freyacodes (approved):
:ok_hand:

PR #106, opened 2018-05-21
Add Histogram for GC pauses

schnapster:
Lavaplayers own warnings are not that accurate, we get many useless warnings in sentry:
image

This PR makes these optional (default behaviour untouched), and instead offers a Histogram metric.


freyacodes:
Is it really safe to rely on `com.sun`?

schnapster:
Idk. That part was copy pasta'd from lavaplayer.

freyacodes (approved):
LGTM

PR #110, opened 2018-06-03
just updated lavaclient url

luaugg:
re-pointed URL at the actual github page and not a branch since latest version is compatible with v2 and v3, so no need to point at a branch.

last commit (merge remote-tracking thingy) is just me being super stupid at git so you can ignore it


freyacodes (approved):

freyacodes:
No worries, the two commits were squashed into one

PR #111, opened 2018-06-04
Lavalink Major Version Header

luaugg:


IMPLEMENTATION.md
```diff
@@ -15,6 +15,10 @@ The Java client has support for JDA, but can also be adapted to work with other
 ## Breaking changes v2.0 -> v3.0 
 The response of `/loadtracks` has been completely changed.
 
+## Non-breaking changes v2.0 -> v3.0
```
freyacodes: I don't see why we need an extra header. I would remove this and change "Breaking" to "Significant" in the other two

LavalinkServer/src/main/java/lavalink/server/io/SocketServer.java
```diff
@@ -76,6 +79,13 @@ public void start() {
         super.start();
     }
 
+    @Override
+    public ServerHandshakeBuilder onWebsocketHandshakeReceivedAsServer(WebSocket conn, Draft draft, ClientHandshake request) throws InvalidDataException {
+        ServerHandshakeBuilder builder = super.onWebsocketHandshakeReceivedAsServer(conn, draft, request);
```
freyacodes, line 84: Are you sure super's method is not empty?

freyacodes (changes requested):

LavalinkServer/src/main/java/lavalink/server/io/SocketServer.java
luaugg: yes i'm sure, it returns an empty handshake impl but it seems java websocket initializes it elsewhere. it returns all the necessary websocket headers as defined in the RFC, its own server name as well as the custom header i specified. i also printed out all the headers received in my own client which returned exactly what i described while using the branch here so i'm very sure it works.

image

^ that was the image i posted in the jda server of the initial commit just before i changed Lavalink-Version to Lavalink-Major-Version as per your request, using code in my client that really does just print out the provided headers from the handshake

IMPLEMENTATION.md
luaugg: sure, you want me to just change "Non-breaking" to "Significant" then?

LavalinkServer/src/main/java/lavalink/server/io/SocketServer.java
freyacodes: Ah fair enough, I didn't stop to consider the return value of it

IMPLEMENTATION.md
luaugg: oh no sorry i gotcha, significant changes for v2 -> v3 and v1 -> v2 and then list em all without the need for an extra markdown header

Frederikam: No, remove the header completely, and have it be under the "Breaking changes v2.0 -> v3.0", but rename "Breaking" to "Significant".

freyacodes (approved):

PR #112, opened 2018-06-05
Add SharpLink

kotx:


kotx:
Don't know whether to change `(Discord.Net)` to `(Discord.Net, C#)` because it doesn't support just C#, as far as I know.

luaugg:
Would probably be better if you just put .NET as the “language” so as to be compatible with other langs that can use your lib.

Besides that, great job on the new lib!

devoxin:
We didn't list this because the libs a WIP with a few bugs 🤔

freyacodes:
I would put C# or .NET, at least for the sake of consistency.

kotx:
(do not merge yet, need confirmation from @Devoxin)

devoxin:
I believe SharpLink is now ready for production use so this could probably be merged (once conflicts are fixed)

freyacodes (approved):
:+1:

PR #113, opened 2018-06-12
fix loadtracks structure documentation

luaugg:
was quite misleading before, hopefully this makes it better


luaugg:
please go ahead and merge #114 instead as it's more complete and shows an example too

PR #116, opened 2018-06-16
Load Types

luaugg:
Allows clients to determine if a call to the /loadtracks endpoint resulted in: a successful track load, a playlist/search load, no matches found or a load failure. Previously Lavalink simply returned an empty array of tracks for the last two cases. This PR doesn't change any of that: it still behaves exactly like it did before but with an added `loadType` field that is equal to either `TRACK_LOADED`, `PLAYLIST_LOADED`, `SEARCH_RESULT`, `NO_MATCHES` or `LOAD_FAILED`.

image

Proof that this PR works with search results. ^
Additionally, with the latest commit it might be feasible to remove the `isPlaylist` field and have it be replaced with the `loadType` one, however this would be a breaking change for clients like my own which expect strictly-formatted responses, or those which already have support for that feature.

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -232,7 +232,6 @@ Authorization: youshallnotpass
 Response:
 ```json
 {
-  "isPlaylist": false,
   "loadType": "TRACK_LOADED",
   "playlistInfo": {},
   "tracks": [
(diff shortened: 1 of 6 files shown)


luaugg:
#117 builds onto the commits from this PR since I didn't make a new fork. Feel free to close one or both if need be.

freyacodes (changes requested):
LGTM, but requires documentation and a note in the changelog

freyacodes (approved):

PR #123, opened 2018-06-29
Add Lavalink.NET to Client Libraries

DevYukine:
Adds my own C# written Library for Lavalink to the Client Libraries, works with any C# Discord API Library.


luaugg:
you could possibly change the targeted "language" from "C#" to just ".NET", similar to #112, especially since the name of your library ends with ".NET"

DevYukine:
Alright done that 👍

freyacodes (approved):
Sorry for the delay. This LGTM.

Frederikam:
Why did that review twice :thinking:

PR #125, opened 2018-07-18
Fix a race condition in the AudioLoader

schnapster:
There are conflicts with the v3 branch. Not sure how we would update both branches with the fix (I can resolve them just not sure whether we merge this fix into both, or merge dev into v3 after this PR is merged)?

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/player/AudioLoader.java
+++ b/LavalinkServer/src/main/java/lavalink/server/player/AudioLoader.java
@@ -31,70 +31,61 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


schnapster:
I totally forgot to give some context:

The AudioLoader can lock up, when lavaplayer's `loadItem` returns (and therefore notifies) a result before the `loadSync` method has reached the line where it waits. The thread will then wait forever there.

Here's a stack of a waiting thread: https://wastebin.party/detatipocu.css
The issue can be easily reproduced after a few curls to the `loadtracks` endpoint by using identifiers/loaders that return very fast, like looking up a local audio file.

freyacodes (approved):
LGTM

schnapster:
v3 merge done in 0adc5c879e80687fa40cac0b32c3c6a33aef1838

PR #133, opened 2018-07-23
Create a changelog file for v2+ releases

schnapster:


CHANGELOG.md
```diff
@@ -0,0 +1,56 @@
+# Changelog
+
+Each release usually includes various fixes and improvements.
+The most noteworthy of these, as well as any features and breaking changes, are listed here.
+
+## v3.0
+* **Breaking:** The minimum required Java version to run the server is now Java 10.   
+**Please note**: Java 10 will be obsolete
+as of [September 2018 with the release of Java 11](http://www.java-countdown.xyz/). Expect a Lavalink minor version release that will be targetting
```
freyacodes: This is a breaking change that warrants a new major version

schnapster: I disagree. It is not a breaking change for users of the docker images, for example.
I personally don't consider the forseeable and necessary bump to the next Java LTS a breaking change of Lavalink's public API - an argument could even be made that it is a patch as all it is doing is a replacing a then outdated version of Java with the current one.
Any user that is running Java applications on a Java 10 JVM and does not update to a Java 11 JVM shortly after its release it running outdated and unsupported software. If someone is seriously planning to keep running applications on Java 10 JVMs after they are deprecated, they are on their own to deal with the consequences of a bad and wrong decision.
If a user can't adhere to the commitment of Oracle's new release schedule, they should not update to Java 9+ in the first place.
A user that updates Lavalink v3.0 to Lavalink v3.x that requires Java 11, but does not update their unsupported and obsolete JVM version to a supported one is doing something wrong imho.

Now, obviously, you are right saying that this a "breaking" change, as in, things can break if users behaviour is incorrect. That's why we do our best to inform users of this expected future change - by including it in the changelog at the very same place where users have to make the decision whether they want to start using non-LTS Java.

schnapster: Decided after a short chat that there is room for interpretation here. The safe choice is to bump the major version, so let's roll with that.

freyacodes (approved):

PR #137, opened 2018-08-14
Add versioning policy

schnapster:
Opinions from client devs and users very much welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,39 @@ Being used in production by FredBoat, Dyno, Rythm, LewdBot, and more.
 
 Please see [here](CHANGELOG.md)
 
+## Versioning policy
+
+The public api ("api" in a very broad sense) of Lavalink can be categorized into two main domains:
+- **Client Domain:** The api exposed to clients, consisting of both the websocket protocol, and any public http endpoints
+- **Server Domain:** The server application with its runtime environment, its configuration, etc.
+
+Changes that might be breaking to one domain need not be breaking to the other.
+Examples: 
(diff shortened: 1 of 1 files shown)


freyacodes (approved):

freyacodes:
>Opinions from client devs and users very much welcome.

Keeping this open for comments

luaugg:
this sounds like a good idea and avoids the issue of making a brand new major release for changes as simple as adding `throws` to one method.

since lavalink is stable and very production-ready, would the versions start from `1.x.x.x`? since `0` is commonly used to indicate an unstable, beta/alpha release.

schnapster:
> avoids the issue of making a brand new major release for changes as simple as adding throws to one method

It does not. There is still a major version. But it avoids bumping the api version for such changes (or rather similar ones. The Lavalink server is not a library and does not have to worry about changing method contracts, but it has analogical concerns in its other domains).

>since lavalink is stable and very production-ready, would the versions start from 1.x.x.x? since 0 is commonly used to indicate an unstable, beta/alpha release.

The v1 -> v2 and v2-> v3 bumps happened mostly due to changes in the client api. I would therefore say picking up where Lavalink is right now avoids the most confusion, meaning: The current api version shall be 3. There haven't been any other releases yet, so now is a good time to adopt a better defined versioning policy with a minimum of changes / mental gymnastics required.

freyacodes:
The next release will be a breaking configuration change that doesn't affect the API, thus it will be released as `v3.1`

PR #143, opened 2018-09-05
Update dependencies (experimental/magma)

shikhir-arora:
Mainly for magma 0.5.0 and jaap 1.3 accordingly but updated the rest. Shouldn't be any issues.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,7 +1,7 @@
 buildscript {
     ext {
-        springBootVersion = '2.0.2.RELEASE'
-        gradleGitVersion = '1.5.1'
+        springBootVersion = '2.0.4.RELEASE'
+        gradleGitVersion = '1.5.2'
         sonarqubeVersion = '2.6.2'
     }
     repositories {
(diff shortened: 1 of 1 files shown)


schnapster:
Not sure if bumping logback is a good idea here. I think there were some breaking issues with this version unless slf4j is bumped too.

schnapster:
See here: https://logback.qos.ch/news.html

> The 1.3.x series is Jigsaw/Java 9 modularized and requires slf4j-api version 1.8.x.

shikhir-arora:
FWIW, we've been running this for ~2 days with no noticeable problems, using OpenJDK 11 (see screenshot). Since we're developing I have the LL log settings set to `trace` (root still set at `info` to avoid the mess 😆) which lets us see all the socket connections and events, and a quick look-through shows nothing breaking/not as it should be, and Sentry looks good as well.

carbon_2018-09-06_20-03-41

freyacodes (approved):

schnapster:
Thanks (especially for the testing)!

PR #147, opened 2018-09-30
Release v3.1

freyacodes:
* Replaced JDAA with Magma
* Added an event for when the Discord voice WebSocket is closed
* Replaced Tomcat and Java_Websocket with Undertow. WS and REST is now handled by the same server and port. Port is specified by `server.port`.

Further testing is appreciated.


schnapster:
You can set magma to 0.6.1 here.

freyacodes:
Ah true, you merged the PR

shikhir-arora:
You could also set LP to @napstr's LP fix on Jitpack for the OPUS tags 👀

schnapster:
I would prefer to avoid that, as my build has weird side effects of missing natives.

shikhir-arora:
@napstr Ah okay, makes sense.

GigaFyde:
In that case, could I interest you in this then? https://github.com/GigaFyde/lavaplayer-main

It's a highly stripped down version of lavaplayer which has been working like a charm for me for the past few days.
It has the Twitch fix, Artwork pr, and also the opus tag fix

And so far no side effects or errors

freyacodes:
@GigaFyde FYI your repo violates the license https://github.com/sedmelluq/lavaplayer/blob/master/LICENSE

GigaFyde:
My bad, completely forgot about the license, time to fix that

GigaFyde:
Should be fixed now 😄

freyacodes:
If we are going to switch to this alternative lavaplayer, I would do so in a separate release.

PR #150, opened 2018-10-08
Allow for resuming closed websockets

freyacodes:
This PR will make it possible to resume a Lavalink session after it has been closed. This may be particularly useful if you lose access to the lavalink server for a moment by accident. When resuming is enabled and the socket is closed, lavalink will keep sending audio until a timeout is reached.

Todo:

- [x] [Document resuming](https://github.com/Frederikam/Lavalink/blob/feature/resume/IMPLEMENTATION.md)
- [x] iImplement `configureResuming`
- [x] Implement resuming
- [x] ~~Implement `reqState`~~
- [x] ~~Implement `resState`~~
- [x] Replay events on resume
- [x] Support this in LLC for testing

Feedback is welcomed.


freyacodes:
Currently, Magma does not have an API to check if we are still connected to one of Discord's voice servers. I am thinking the `resState` for whether or not we are connected may be redundant, as the voice state is can be used for checking if we are *supposed* to be connected.

It might be worth testing if it is viable to exclude the `connected` property in the JSON response. Otherwise, I will need to create a Magma PR.

schnapster:
Does this help?
https://github.com/napstr/Magma/pull/15

freyacodes:
@napstr thanks! API looks good to me. Would you say we should provide the raw eval name in the Lavalink API?

schnapster:
Sorry I'm a bit out of touch with this PR. What do mean by "raw eval name" ?

freyacodes:
@napstr for the sake of resuming, the `resState` and `reqState` messages are introduced to tell the client what the state of the player is, and if voice is still connected.

The info for the voice connection was initially meant to be represented as a boolean, however we could instead return the `WebsocketConnectionState.Phase` ~~eval~~ ***enum*** name.

schnapster:
Should be fine as long as the magma namespace doesn't leak out. In case another state is added in the future, make sure to gracefully handle a default case (or similar).

freyacodes:
@napstr to be honest, I think the way I'll interpret it in my code is something like
```java
boolean connected = phase == "CONNECTED";
```

freyacodes:
In hindsight, I don't really like the reqState/resState ops. I'm thinking it might be more sensible just to replay events like Discord does. That makes it possible to track which audio tracks were ended in the meanwhile (if at all) and why.

freyacodes:
The server now replays events instead of having to request player/magma state from the server. This simplifies things quite a bit and makes it much easier to implement on clients. You should only have to facilitate resuming itself, which is fairly easy.

Yucked:
Any ETA?

freyacodes:
@Yucked ETA for what?

Server support has been implemented, but it hasn't been implemented into a real bot yet.

Yucked:
Oh I thought the resuming wouldn't be implemented till this PR was closed or something

freyacodes:
A release candidate is out https://github.com/Frederikam/Lavalink/releases/tag/3.2-rc1

Yucked:
@Frederikam 

Will I've to configure resume for each node or per Lavalink session?

freyacodes:
@Yucked I'm not completely sure what you mean, but anything you send the server via WS is specific to that session.

PR #154, opened 2018-10-26
Equalizer stuff

devoxin:
Alright so here's the (fixed) pull request for equalizer functionality within Lavalink. This has been tested and worked flawlessly for the duration of my testing :)

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -124,7 +124,9 @@ Using the player equalizer
 }
 ```
 There are 16 bands (0-15) that can be changed.
-`gain` must be a float between -0.25 and 1.0, where 0.0 indicates no change.
+`gain` is the multiplier for the given band. The default value is 0. Valid values range from -0.25 to 1.0,
+where -0.25 means the given band is completely muted, and 0.25 means it is doubled. Modifying the gain could
+also change the volume of the output.
 
 Tell the server to potentially disconnect from the voice server and potentially remove the player with all its data.
 This is useful if you want to move to a new node for a voice connection. Calling this op does not affect voice state,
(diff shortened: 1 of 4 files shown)


IMPLEMENTATION.md
```diff
@@ -110,6 +110,22 @@ Set player volume. Volume may range from 0 to 1000. 100 is default.
...
+    "bands": [
+        {
+            "band": 0,
+            "gain": 0.2
+        }
+    ]
+}
+```
+There are 16 bands (0-15) that can be changed.
+`gain` must be a float between -0.25 and 1.0, where 0.0 indicates no change.
```
schnapster: nit: saying "must" implies other values could be not accepted possibly even throwing an error or other weird things happening, what actually happens is that lavaplayer silently caps them at the values you mentioned. Would be nice to have it worded in the docs that way. Also, to answer the question of the efffects, maybe just copy over or link the lavaplayer docs:
```
  /**
   * @param band The index of the band. If this is not a valid band index, the method has no effect.
   * @param value The multiplier for this band. Default value is 0. Valid values are from -0.25 to 1. -0.25 means that
   *              the given frequency is completely muted and 0.25 means it is doubled. Note that this may change the
   *              volume of the output.
   */
```
https://github.com/sedmelluq/lavaplayer/blob/master/main/src/main/java/com/sedmelluq/discord/lavaplayer/filter/equalizer/EqualizerConfiguration.java#L19

freyacodes (approved):
I have never had time to look into the new equalizer in Lavaplayer, but the code looks fine. I trust this has been tested exhaustively?

devoxin:
The equalizer integration is based on code from my Java/Kotlin bot, as for everything else I tested it locally using Lavalink.py and encountered no issues at all 👍

PR #159, opened 2018-11-09
Add Lavalink-Major-Version

### LINKED ISSUE #142: Send the Lavalink-Major-Version header in REST responses
As per docs: 

- Lavalink v3.0 now reports its version as a handshake response header. Lavalink-Major-Version has a value of 3 for v3.0 only. It's missing for any older version.

This would be a very useful feature, if not for the fact that most WS libraries do not expose the handshake, and thus provide no ability to recover that header. While at present it's possible to distinguish Lavalink 3.x and 2.x by checking REST response type (array vs object), this is a fragile approach that might not work in the future.

devoxin:
Fixes #142 
headers

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -307,6 +307,8 @@ Additionally, in every `/loadtracks` response, a `loadType` property is returned
 * `NO_MATCHES` - Returned if no matches/sources could be found for a given identifier.
 * `LOAD_FAILED` - Returned if Lavaplayer failed to load something for some reason.
 
+All REST responses from Lavalink include a `Lavalink-Api-Version` header.
+
 ### Special notes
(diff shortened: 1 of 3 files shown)


freyacodes:
That header is misleading. It shows the API version, not the major version.

I know the same currently goes for the WS headers, but I don't want that to proliferate.

devoxin:
Changed it to Api

freyacodes:
For the sake of consistency, this header should be sent with each response. As far as I know, this can be accomplished with [a filter](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/filter/OncePerRequestFilter.html).

Additionally, this change should be described in IMPLEMENTATION.md

devoxin:
Should be good now

freyacodes (approved):

PR #164, opened 2019-01-16
Compatibility with Java 8 through 10

MinnDevelopment:
Simply had to change a few syntax issues and fork magma, now this supports java 8.

I'm hosting my fork of magma on [jcenter](https://bintray.com/minndevelopment/maven/magma) since I find jitpack annoying to use for production due to constant issues with timeouts.

I haven't really tested this on a real project as I'm missing the setup so I'd recommend to do this before merging. I was able to compile properly with JDK 8 though.


freyacodes:
I don't use Java 8 or 9, but a build is available here for those that do: https://ci.fredboat.com/repository/download/Lavalink_Build/6275:id/Lavalink.jar?guest=1

Issue #163, opened 2018-12-30
REST API: /loadtracks return friendlyexception


zaza7:
When using the REST API endpoint /loadtracks and it fails due to copyright block, the response given back is the following
 ```
{
    "playlistInfo": {},
    "loadType": "LOAD_FAILED",
    "tracks": []
}
```
is it possible to also have the reason why the track failed to load? For example "This video contains content from *whomever*, who has blocked it in your country on copyright grounds."

freyacodes:
That should be possible. The exception is provided here: https://github.com/Frederikam/Lavalink/blob/master/LavalinkServer/src/main/java/lavalink/server/player/AudioLoader.java#L101

schnapster:
Implemented by #174 (Thanks @Devoxin!)

PR #174, opened 2019-03-02
Return FriendlyException message on LOAD_FAILED.

devoxin:
This PR aims to add in the functionality mentioned in #163. I also cleaned up how authorization is handled by using a Pre-Request filter, which removes the need to check authorization in the code of every endpoint.

When a request yields LOAD_FAILED, `exception` is returned in the response object. The other keys exist for the purpose of ensuring this is backwards-compatible change.

I'm not 100% sure on the implementation of how the exception is returned, so feedback is very much appreciated.


LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.java
```diff
@@ -39,15 +41,18 @@
...
+import org.springframework.web.servlet.config.annotation.InterceptorRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 
 import javax.servlet.http.HttpServletRequest;
 import java.io.IOException;
 import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
 
+@Configuration
```
schnapster: Setting a "RequestAuthorizationFilter" has nothing to do with this specific RestHandler. It will also be applied to all endpoints, and is therefore highly misleading if it exists in this class. Probably best to be put into a standalone config class in the `lavalink.server.config` package.

LavalinkServer/src/main/java/lavalink/server/io/RequestAuthorizationFilter.java
```diff
@@ -0,0 +1,32 @@
...
+
+    public RequestAuthorizationFilter(ServerConfig config) {
+        this.serverConfig = config;
+    }
+
+    @Override
+    public boolean preHandle(HttpServletRequest request, HttpServletResponse response, Object handler) throws Exception {
+        String authorization = request.getHeader("Authorization");
+
+        if (authorization == null || !authorization.equals(serverConfig.getPassword())) {
```
schnapster: This looks like it will break the debug endpoint which relies in a different password.

schnapster (changes requested):

schnapster:
Also this is missing an update of https://github.com/Frederikam/Lavalink/blob/master/IMPLEMENTATION.md

LavalinkServer/src/main/java/lavalink/server/io/RequestAuthorizationFilter.java
freyacodes: The debug endpoint should probably be removed

schnapster:
Personally I find the usefulness of a random plain string exception close to zero. Could you add an explain of the value of this PR?

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.java
devoxin: Perhaps I'm just tired, but I'm not exactly sure what you mean?

devoxin:
A load can fail when, for example, a video is unavailable due to region-blocking by copyright. I believe letting clients display a reason for not being able to load a track offers a slightly better user experience than a generic error of "Track couldn't be loaded", or nothing at all.

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.java
schnapster: A mix of Separation of Concerns and the Spring Boot docs.

schnapster:
Displaying arbitrary strings that you receive from some api to endusers is a bad idea.

freyacodes:
@napstr I don't understand where you are coming from. Lots of protocols encourage that, such as with HTTP status strings and WS close reasons.

Edit: Here I was thinking about it for logging purposes, not for UX.

devoxin:
I disagree. In cases where the video is from YouTube, Lavaplayer scrapes the exact reason YouTube provides to the user when viewed in an actual browser, so the messages should still be understandable to end users. Additionally I display such messages in my bot to users and have had 0 issues with it.

It's probably only a small thing but I think in cases where users try to play a certain track and it fails to load, a generic error of "Unable to load video" will probably make the user join the support server and ask why it failed, only to receive a canned response consisting of a list of reasons as to why it could've failed.

schnapster:
My bad. I was under the impression these would by any kind of exceptions. However it is just the `FriendlyException`, whose class docs say:
>@param friendlyMessage A message which is understandable to end-users

schnapster:
What do you think about including the [severity](https://github.com/sedmelluq/lavaplayer/blob/master/main/src/main/java/com/sedmelluq/discord/lavaplayer/tools/FriendlyException.java#L10)?

```json
  "loadType": "LOAD_FAILED",
  "playlistInfo": {},
  "tracks": [],
  "exception": {
    "friendlyMessage": "The uploader has not made this video available in your country.",
    "severity": "COMMON"
  }
```

devoxin:
Not too sure about that one, might be useful for developers as an indicator of how severe the exception is but other than that it's probably a little useless, although I've no objections to including that in the response.

IMPLEMENTATION.md
```diff
@@ -309,6 +309,20 @@ Additionally, in every `/loadtracks` response, a `loadType` property is returned
...
 
+If the loadType is `LOAD_FAILED`, the response will contain an `exception` which provides an insight as to why loading of the track failed.
+
+```json
+{
+  "loadType": "LOAD_FAILED",
+  "playlistInfo": {},
+  "tracks": [],
+  "exception": {
+    "friendlyMessage": "The uploader has not made this video available in your country.",
```
freyacodes: This should probably just be `message`

IMPLEMENTATION.md
```diff
@@ -309,6 +309,20 @@ Additionally, in every `/loadtracks` response, a `loadType` property is returned
...
+    "severity": "COMMON"
```
freyacodes, line 323: Severity should be explained

LavalinkServer/src/main/java/lavalink/server/io/RequestAuthorizationFilter.java
```diff
@@ -0,0 +1,32 @@
...
+            log.warn("Authorization failed");
```
freyacodes: This should probably log the IP and request path

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.java
```diff
@@ -47,7 +51,7 @@
 import java.util.concurrent.CompletionStage;
 
 @RestController
-public class AudioLoaderRestHandler {
+public class AudioLoaderRestHandler implements WebMvcConfigurer {
```
freyacodes: `AudioLoaderRestHandler` affects ALL RestControllers. It is misleading to do the configuration from this class, and it should be done by an `@Configuration` class in the config package.

LavalinkServer/src/main/java/lavalink/server/player/LoadResult.java
```diff
@@ -1,23 +1,28 @@
...
     public String playlistName;
-    public ResultStatus loadResultType;
     public Integer selectedTrack;
+    public FriendlyException exception;
 
-    public LoadResult(List<AudioTrack> tracks, @Nullable String playlistName, ResultStatus loadResultType,
-                      @Nullable Integer selectedTrack) {
+    public LoadResult(ResultStatus loadResultType, List<AudioTrack> tracks,
+                      @Nullable String playlistName, @Nullable Integer selectedTrack,
+                      @Nullable FriendlyException exception) {
```
freyacodes: `exception` is mutually exclusive to all other parameters besides `loadResultType`, so a separate constructor is preferred.

freyacodes (changes requested):

IMPLEMENTATION.md
freyacodes: You might want to otherwise document that the message is okay to show end-users

schnapster (approved):

freyacodes (approved):

PR #199, opened 2019-06-12
Add Shoukaku to Client Libraries

Deivu:
Shoukaku is a Discord.js lavalink client purely made for Discord.js v12


freyacodes:
I would like to put this on hold as long as your library is still in alpha

Deivu:
The library is now usable. Documentation and example codes are also now provided.

freyacodes (approved):

PR #201, opened 2019-06-14
Add Friendly exception to TrackExceptionEvent

### LINKED ISSUE #198: Return FriendlyException from onTrackException
I want to propose to return the FriendlyException that is passed to [onTrackException][trackex] so that the java client can make use of this and have the correct severity instead of it always being [SUSPICIOUS][currentsend].
This currently already is a feature for the `/loadtracks` endpoint but not for the WebSocket connection.

I would like to make a pull request to implement this, but before I do I want to know what people think about this (possibly breaking) change.

[trackex]: https://github.com/Frederikam/Lavalink/blob/dev/LavalinkServer/src/main/java/lavalink/server/player/EventEmitter.java#L69
[currentsend]: https://github.com/FredBoat/Lavalink-Client/blob/master/src/main/java/lavalink/client/player/event/AudioEventAdapterWrapped.java#L57

duncte123:
This PR adds the FriendlyException from lava player to the TrackExceptionEvent so the java client can make use of this and return the correct severity.

Example output: (tested with the lofi hip hop radio stream)
```json
{
  "exception": {
    "severity": "FAULT",
    "cause": "org.apache.http.ConnectionClosedException: Premature end of chunk coded message body: closing chunk expected",
    "message": "Something went wrong when decoding the track."
  },
  "op": "event",
  "type": "TrackExceptionEvent",
  "track": "QAAAkAIALGxvZmkgaGlwIGhvcCByYWRpbyAtIGJlYXRzIHRvIHJlbGF4L3N0dWR5IHRvAApDaGlsbGVkQ293f/////////8AC2hIVzFvWTI2a3hRAQEAK2h0dHBzOi8vd3d3LnlvdXR1YmUuY29tL3dhdGNoP3Y9aEhXMW9ZMjZreFEAB3lvdXR1YmUAAAAAAAAX/A==",
  "error": "Something went wrong when decoding the track.",
  "guildId": "191245668617158656"
}
``` 

Closes #198

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -12,7 +12,7 @@ The Java client has support for JDA, but can also be adapted to work with other 
     * Hixie 76
     * Hixie 75
 
-## Significant changes v3.0 -> v3.1
+## Significant changes v3.0 -> v4.0
 The `error` string on the `TrackExceptionEvent` has been deprecated and replaced by 
 the `exception` object following the same structure as the `LOAD_FAILED` error on [`/loadtracks`](#rest-api)
 
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -12,6 +12,10 @@ The Java client has support for JDA, but can also be adapted to work with other 
     * Hixie 76
     * Hixie 75
 
+## Significant changes v3.0 -> v3.1
+The `error` string on the `TrackExceptionEvent` has been deprecated and replaced by 
+the `exception` object following the same structure as the `LOAD_FAILED` error on [`/loadtracks`](#rest-api)
+
 ## Significant changes v2.0 -> v3.0 
 * The response of `/loadtracks` has been completely changed (again since the initial v3.0 pre-release).
 * Lavalink v3.0 now reports its version as a handshake response header.
@@ -225,7 +229,7 @@ private void handleEvent(JSONObject json) throws IOException {
                     new FriendlyException(
                         jsonEx.getString("message"),
                         FriendlyException.Severity.valueOf(jsonEx.getString("severity")),
-                        new Exception(jsonEx.getString("cause"))
+                        new RuntimeException(jsonEx.getString("cause"))
                     )
             );
             break;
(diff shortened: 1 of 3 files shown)


IMPLEMENTATION.md
```diff
@@ -219,9 +219,14 @@ private void handleEvent(JSONObject json) throws IOException {
...
             break;
         case "TrackExceptionEvent":
+            JSONObject jsonEx = json.getJSONObject("exception");
             event = new TrackExceptionEvent(player,
                     LavalinkUtil.toAudioTrack(json.getString("track")),
-                    new RemoteTrackException(json.getString("error"))
+                    new FriendlyException(
+                        jsonEx.getString("message"),
+                        FriendlyException.Severity.valueOf(jsonEx.getString("severity")),
+                        new Exception(jsonEx.getString("cause"))
```
freyacodes: Should probably be `RuntimeException`

freyacodes (changes requested):
This will have to be documented, particularly with the deprecation of error in the next API version

IMPLEMENTATION.md
```diff
@@ -12,6 +12,10 @@ The Java client has support for JDA, but can also be adapted to work with other
     * Hixie 76
     * Hixie 75
 
+## Significant changes v3.0 -> v3.1
```
freyacodes: As per our policy, this API breaking change would warrant an API bump to v4.0.
```suggestion
## Significant changes v3.0 -> v4.0
```

freyacodes (changes requested):

IMPLEMENTATION.md
duncte123: Mixed up api and major, mb

freyacodes:
Here's what I'll do. I'll create a new branch for you called something like `nextapi` or something. Then I'll merge this PR into that as a v4 branch. 

This will allow me to publish a v4 snapshot version without formally releasing v4.

duncte123:
> Here's what I'll do. I'll create a new branch for you called something like `nextapi` or something. Then I'll merge this PR into that as a v4 branch.
> 
> This will allow me to publish a v4 snapshot version without formally releasing v4.

Sounds good to me

freyacodes (approved):

PR #210, opened 2019-07-28
Add NaN check to stats

NIXC:
On some Mac and Windows machines stats can get a NaN for system processor load, which isn't handled very well by stats causing it to break. This change would just replace that with zero. Its not a perfect fix for stats but at least it doesn't throw errors.

Should fix https://github.com/Frederikam/Lavalink/issues/206


Stonedestroyer:
Fixed by #492

PR #191, opened 2019-05-06
Replace oshi with calls to native JVM apis

schnapster:
Should solve the occasional bugs with oshi, as well as plain wrong and therefore useless CPU readings.

Probably only works on certain JVMs (OpenJDK, IBM/Eclipse J9), and probably not on Windows.

Fixes #158

PR #211, opened 2019-07-30
 Replace oshi with calls to native JVM apis AND check for NaN 

NIXC:
Is basically https://github.com/Frederikam/Lavalink/pull/191 with https://github.com/Frederikam/Lavalink/pull/210 merged on top of it.

Initial results for this working on the mac that was having issues previously are good. The windows 2012 R2 machine from 191 is working as well.

Theoretically my changes should at least be even in terms of compatibility. 

If there is a better way to make this PR please let me know.

### DIFF
--- a/LavalinkServer/build.gradle
+++ b/LavalinkServer/build.gradle
@@ -45,7 +45,6 @@ dependencies {
     compile group: 'org.springframework', name: 'spring-websocket', version: springWebSocketVersion
     compile group: 'ch.qos.logback', name: 'logback-classic', version: logbackVersion
     compile group: 'io.sentry', name: 'sentry-logback', version: sentryLogbackVersion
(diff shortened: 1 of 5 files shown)


freyacodes:
The problem with #191 (and thus also this PR) is that it breaks on some JVMs where OSHI presumably works fine. This is a breaking change I am not happy with.

It would be nice to be able to fall back to the old OSHI system, in case the native method is failing.

schnapster:
> The problem with #191 (and thus also this PR) is that it breaks on some JVMs where OSHI presumably works fine. This is a breaking change I am not happy with.

> It would be nice to be able to fall back to the old OSHI system, in case the native method is failing.

Don't wait for me, I'm sure some of the users of those will contribute the necessary code and do the necessary testing.

PR #238, opened 2019-11-28
Use non-root user for Docker image

james7132:
Root container escapes are common. Upon container escape, the attacker will use the UID of 0 and give root access to the host. This change avoids this possibility by using a non-root user to run the lavalink server.


freyacodes (approved):
This makes sense. Just promise me you are not relying on Docker as a secure jail.

james7132:
No I'm not, but this is just a common sense security precaution.

PR #240, opened 2019-12-08
Update lavaplayer to 1.3.32

duncte123:


freyacodes (approved):
Sorry for the late response! I'm only seeing this now because I was about to do the same thing. Wrong branch though.

duncte123:
> Sorry for the late response! I'm only seeing this now because I was about to do the same thing. Wrong branch though.

Oh whoops, thought that I selected dev

PR #241, opened 2019-12-09
feat: Single APM across all WS connections as default

Xavinlol:


schnapster:
What is the benefit of a singleton APM? Why is that not the default? Just curious. I also think it might be worth documenting that info.

DevYukine:
ImageOne
ImageTwo
ImageThree

We discussed this one on discord already, since a lot of people dont use Java to connect to LL and have multiple ws connections due to scaling of other languages, LL creates one APM for each WS which can lead into a ton of Threads from LP Threadpools & Spamming GC warnings since each APM does GC Warning on its own which also isn't great

Xavinlol:
Basically what @DevYukine said. And i didn't set it to default so it's nothing breaking (created an APM for each WS connection before)

schnapster:
Are there any drawbacks? I would just make it the default (and only) tbh.

freyacodes (approved):

PR #247, opened 2020-01-04
Change Magma build slightly

NIXC:
Adjusts Magma so that it doesn't compile a blank magma folder that contains a manifest in addition to the normal and operational Magma. The folders have the same name but different capitalization, which causes conflicts on Windows if you want to repack the jar for any reason.


freyacodes (approved):
This is even the example given in the Magma readme.

PR #253, opened 2020-02-09
Disable cookie management

Repulser:
Currently testing this PR


ByteAlex:
If it works, make it configurable probably

PR #268, opened 2020-04-07
Bump Lavaplayer to 1.3.46

devoxin:
That time of the year again :P

- Fixes YouTube mix loading.
  - Mix loading is also done in 1 request, so tracks are loaded MUCH quicker than before.
- Fixes YouTube livestream playback.
  - Also adds a retry mechanism so they shouldn't stop after a minute of playback.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ subprojects {
 
     ext {
         //@formatter:off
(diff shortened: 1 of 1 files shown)


schnapster:
LGTM

PR #269, opened 2020-04-10
Replace Magma with Koe

freyacodes:
[Koe](https://github.com/KyokoBot/koe) is a smaller and perhaps more maintainable alternative to Magma.

Koe also relies on JDA-NAS natives for native audio sending.

PR #270, opened 2020-04-11
Update Lavaplayer IP rotator

Thewsomeguy:
Bumping IP rotator version `0.1.7` -> `0.2.3`
I have been running it in prod for over a week and it works fine.


freyacodes:
I will probably test this together with #269

Issue #274, opened 2020-04-18
YouTube: Connection timed out


Axiol:
## Description
Sometimes when I try to play something though a Red bot, there's a bit of time out and then I get an error where the bot returns this

```
Lavalink API returned an unsupported response, Please report it.Query: http://localhost:2333/loadtracks?identifier=https%3A//www.youtube.com/watch%3Fv%3DyzC4hFK5P3g{'timestamp': '2020-04-18T21:34:55.691+0000', 'status': 503, 'error': 'Service Unavailable', 'message': 'Service Unavailable', 'path': '/loadtracks', 'query': 'https://www.youtube.com/watch?v=yzC4hFK5P3g', 'encodedquery': 'http://localhost:2333/loadtracks?identifier=https%3A//www.youtube.com/watch%3Fv%3DyzC4hFK5P3g', 'loadType': <LoadType.LOAD_FAILED: 'LOAD_FAILED'>}Query: http://localhost:2333/loadtracks?identifier=https%3A//www.youtube.com/watch%3Fv%3DyzC4hFK5P3g{'timestamp': '2020-04-18T21:40:43.726+0000', 'status': 503, 'error': 'Service Unavailable', 'message': 'Service Unavailable', 'path': '/loadtracks', 'query': 'https://www.youtube.com/watch?v=yzC4hFK5P3g', 'encodedquery': 'http://localhost:2333/loadtracks?identifier=https%3A//www.youtube.com/watch%3Fv%3DyzC4hFK5P3g', 'loadType': <LoadType.LOAD_FAILED: 'LOAD_FAILED'>}
```

## Version info
**Client used:**

Output of `java -jar Lavalink.jar --version`:
```
Version:        e270ec688cbad900d9de292a9e615f15c48a9a82-SNAPSHOT
Build:          986
Build time:     23.03.2020 20:28:53 UTC
Branch          BullyJava11
Commit:         e270ec6
Commit time:    23.03.2020 19:56:15 UTC
JVM:            13.0.2
Lavaplayer      1.3.38
```

```
2020-04-18 23:23:04.246  INFO 82 --- [main] lavalink.server.Launcher                 : Starting Launcher on 8cdd9ea71b96 with PID 82 (/data/cogs/Audio/Lavalink.jar started by docker in /data/cogs/Audio)
2020-04-18 23:23:04.252  INFO 82 --- [main] lavalink.server.Launcher                 : No active profile set, falling back to default profiles: default
[...]

PR #276, opened 2020-04-20
Allow setting YouTube timeout

freyacodes:
Possibly a workaround for #274, unless there is a different underlying cause

### DIFF
--- a/LavalinkServer/application.yml.example
+++ b/LavalinkServer/application.yml.example
@@ -16,6 +16,7 @@ lavalink:
     bufferDurationMs: 400
     youtubePlaylistLoadLimit: 6 # Number of pages at 100 each
     youtubeSearchEnabled: true
+    youtubeTimeout: -1
     soundcloudSearchEnabled: true
     gc-warnings: true
     #ratelimit:
--- a/LavalinkServer/src/main/java/lavalink/server/config/AudioPlayerConfiguration.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/config/AudioPlayerConfiguration.kt
@@ -22,12 +22,12 @@ import com.sedmelluq.lava.extensions.youtuberotator.planner.RotatingIpRoutePlann
 import com.sedmelluq.lava.extensions.youtuberotator.planner.RotatingNanoIpRoutePlanner
 import com.sedmelluq.lava.extensions.youtuberotator.tools.ip.Ipv4Block
 import com.sedmelluq.lava.extensions.youtuberotator.tools.ip.Ipv6Block
+import org.apache.http.client.config.RequestConfig
 import org.slf4j.LoggerFactory
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import java.net.InetAddress
 import java.util.function.Predicate
-import java.util.function.Supplier
 
 /**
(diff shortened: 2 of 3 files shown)


cloudrac3r:
This helps me!

Are you able to add a property for the timeout for generic HTTP(S) streams as well? This would be extremely useful for me. I have a system where sometimes I pass Lavalink the YouTube stream URL directly instead of using the LavaPlayer loader. These stream URLs time out after the classic 3000 ms because Lavalink doesn't consider them to be YouTube. I'd like to adjust that timing too.

freyacodes:
@cloudrac3r I am a bit stumped by your request. Generic YouTube stream URLs?

In Lavaplayer, the YouTube source provider gets asked if a given URL corresponds to it. The generic HTTP loader (which doesn't work for YouTube!) gets asked as one of the last source providers. If the HTTP provider truly tries to load it, it should be evident by the stack trace.

Though that doesn't explain your problem. The track loader API that Lavalink uses to load Lavaplayer tracks is the same that we use for playing tracks. In theory, we use the same HTTP client in both cases.

cloudrac3r:
Hmm. Can you tell me exactly which kinds of outgoing requests are affected by the youtubeTimeout setting?

freyacodes:
Should be anything sent by `YoutubeAudioSourceManager`, which should be anything to youtube.com.

Could you post your stack trace?

cloudrac3r:
The urls in question are for `anything.googlevideo.com`. I don't have logs or a stack trace of this at the moment, but I'll be sure to provide them when I next notice this happening.

freyacodes:
Now that makes more sense. I'm surprised you're getting a timeout on whatever cache is near you. Some large media providers like YouTube and Netflix have caching servers located with local ISPs.

We can change the timeout of the HTTP manager, though I don't particularly like the idea of using special config parameters like this. Perhaps a more generic config format would be nicer. Though I am not sure what that would look like.

cloudrac3r:
I tried the specific googlevideo URL on a few machines in different regions and IIRC it was slow in all of them. Just YouTube being YouTube. Next time I'm affected I will do a proper investigation and write down my findings for you.

This PR in its current form is definitely an improvement for normal use, though.

cloudrac3r:
I haven't had issues with timeouts since I last commented.

MrRazamataz:
Hey @freyacodes , I have a similar issue.
I have attempted to setup ipv6 addresses using your guide (although I am using a /64 subset from hetzner, I did also try the /48 on a VPS), and I seem to be getting a timed out error.
Sorry if this is a dumb question, but where can I try out this build? I tried the Dev build from the CL but it looks like it's not been merged yet, I also tried to build this repo with the changes, but was unsuccessful. 

My error is here:
https://cdn.mrrazamataz.ga/data/lavalinkipv6%20error.txt

Thank you.

freyacodes:
The problem is with your network configuration. Setting a higher timeout will not help.

MrRazamataz:
> The problem is with your network configuration. Setting a higher timeout will not help.

Would you have any idea what I have done wrong? I have read your guide and the only bit I didnt do was the getting a /48 block as I had a /64 set that came with my machine from hetzner. Would you know any possible cause/fix?

Yes, my ping6 tests are successful.

PR #297, opened 2020-05-22
fix: bump lavaplayer

favna:
There don't seem to be any breaking change for Lavalink in the versions in between, however there are some important fixes. In particular in 1.3.43+1.3.44 Soundcloud fixes, 1.3.45 and 1.3.47 YouTube live streams, 1.3.46 Youtube mixes, and in 1.3.49 Twitch playback.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ subprojects {
 
     ext {
         //@formatter:off
-        lavaplayerVersion               = '1.3.38'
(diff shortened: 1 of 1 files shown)


OtterBoops:
This is already on dev. Use that branch instead.

favna:
Oh woops I didn't check that branch. Closing then.

freyacodes:
I will draft a release of the dev version whenever I get around to it.

PR #315, opened 2020-06-18
Add support for endTime

### LINKED ISSUE #179: Make endTime work
> Should probably make a point that, while `endTime` is documented, it doesn't affect anything; i.e., there's no actual code that does anything with it if received.

_Originally posted by @Devoxin in https://github.com/Frederikam/Lavalink/issues/170#issuecomment-473971334_

sh30801:
This PR fixes #179, and makes end time functional.

The only thing I am not sure about is handling the seek portion. Right now it will end the track if the seek position is greater than the endtime position, but perhaps it should just cancel the marker and keep the track playing.

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
@@ -48,8 +48,8 @@ class WebSocketHandlers(private val contextMap: Map<String, SocketContext>) {
             player.setVolume(json.getInt("volume"))
         }
 
-        if (json.has("stopTime")) {
-            val stopTime = json.getLong("stopTime")
+        if (json.has("endTime")) {
+            val stopTime = json.getLong("endTime")
             if (stopTime > 0) {
                 val handler = TrackEndMarkerHandler(player)
                 val marker = TrackMarker(stopTime, handler)
--- a/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
@@ -49,7 +49,6 @@ class WebSocketHandlers(private val contextMap: Map<String, SocketContext>) {
         }
 
         if (json.has("stopTime")) {
-
             val stopTime = json.getLong("stopTime")
             if (stopTime > 0) {
                 val handler = TrackEndMarkerHandler(player)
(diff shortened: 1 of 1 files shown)


LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
```diff
@@ -47,6 +48,16 @@ class WebSocketHandlers(private val contextMap: Map<String, SocketContext>) {
             player.setVolume(json.getInt("volume"))
         }
 
+        if (json.has("stopTime")) {
```
devoxin: As per the [IMPLEMENTATION.md](https://github.com/Frederikam/Lavalink/blob/master/IMPLEMENTATION.md#outgoing-messages), this should be `endTime`.

LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
```diff
@@ -47,6 +48,16 @@ class WebSocketHandlers(private val contextMap: Map<String, SocketContext>) {
...
+
+            val stopTime = json.getLong("stopTime")
```
devoxin: See above comment.

devoxin (changes requested):

devoxin:
> The only thing I am not sure about is handling the seek portion. Right now it will end the track if the seek position is greater than the endtime position, but perhaps it should just cancel the marker and keep the track playing

Lavaplayer automatically handles this. When a seek is requested that exceeds the marker position, the marker handler will be invoked with a state denoting so. I believe this should already be handled [by the `BYPASSED` check](https://github.com/Frederikam/Lavalink/blob/master/LavalinkServer/src/main/java/lavalink/server/player/TrackEndMarkerHandler.java#L37), where `BYPASSED` is the state provided when the marker position is exceeded, as [denoted here](https://github.com/sedmelluq/lavaplayer/blob/master/main/src/main/java/com/sedmelluq/discord/lavaplayer/track/TrackMarkerHandler.java#L29).

My only concern is the MarkerHandler line linked above (L37).
`if (state.equals(MarkerState.REACHED) | state.equals(MarkerState.BYPASSED))` should perhaps be `if (state.equals(MarkerState.REACHED) || state.equals(MarkerState.BYPASSED))` instead.

freyacodes (commented):
The code looks fine, but has this been tested?

sh30801:
> 
> 
> The code looks fine, but has this been tested?

Yes it has. It works fine both with and without and endtime. Passing in an endtime ends it as expected.

freyacodes (approved):
:ok_hand:

PR #326, opened 2020-07-06
Add playerUpdateInterval option to config

TheEssem:
This PR adds an option to the config (`playerUpdateInterval`) to set the interval for when a playerUpdate WS event is sent, like so:
```
playerUpdateInterval: 3
```
This is useful for cases when, for example, you want to get the current song duration every second or you rarely want to receive said events. As a bit of a side effect, it also adds `serverConfig` as an option to Player which might not be the most optimal solution, so suggestions are appreciated.

### DIFF
--- a/LavalinkServer/application.yml.example
+++ b/LavalinkServer/application.yml.example
@@ -15,6 +15,7 @@ lavalink:
       local: false
     bufferDurationMs: 400
     youtubePlaylistLoadLimit: 6 # Number of pages at 100 each
+    playerUpdateInterval: 5
     youtubeSearchEnabled: true
     soundcloudSearchEnabled: true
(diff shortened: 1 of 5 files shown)


LavalinkServer/application.yml.example
```diff
@@ -15,6 +15,7 @@ lavalink:
       local: false
     bufferDurationMs: 400
     youtubePlaylistLoadLimit: 6 # Number of pages at 100 each
+    playerUpdateInterval: 5
```
freyacodes:
```suggestion
    playerUpdateInterval: 5 # How frequently to send player updates to clients, in seconds
```

freyacodes (approved):
Looks great. Just want to add this comment

PR #336, opened 2020-08-04
Update Koe

Dramex:
https://github.com/KyokoBot/koe/releases/tag/1.0.1


freyacodes (approved):
Fix koe using wrong port

PR #358, opened 2020-10-04
Update LP

devoxin:
<https://github.com/Devoxin/lavaplayer/releases/tag/1.3.54.3>
- Fixes age-restricted track playback by checking for more player config formats.
  - Less noteworthy, but HTML of tracks w/o a matching player config is now logged to debug.


freyacodes:
Is it only playback that is broken and not the track loading?

devoxin:
As far as I could gather, yes.

(Screenshot taken of playback attempts pre-fix. It would load fine but break randomly. A final fix was pushed to correct this).

freyacodes (approved):

PR #359, opened 2020-10-04
bump lavaplayer version to fix age restricted videos playback

sebm253:


freyacodes:
You're 3 minutes too slow.

#358

PR #372, opened 2020-10-24
Update LP Version

applebee1558:
Bump lavaplayer to newer version which fixes many bugs when loading tracks.


freyacodes:
Have you considered using the dev branch, which has this version?

applebee1558:
@Frederikam yes, but shouldn’t we keep master updated so anyone that comes doesn’t have to go off a sub-branch just to get things working?

freyacodes:
I've decided to merge dev instead, so you should be good now.

PR #264, opened 2020-03-30
Working filter stuff

devoxin:

PR #398, opened 2020-11-19
Working filters with soundcloud on 1.3.60

DaliborTrampota:


freyacodes:
Closing as the code is around 250 commits behind. See #264 which I've just now updated

PR #417, opened 2020-12-19
Add rotation and distortsion filters

Mandruyd:
Added the distortion effect from lavadsp.
Added the new effect added in the latest lavadsp version, rotation.

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -176,7 +176,25 @@ JSON comments are for illustration purposes only, and will not be accepted by th
     vibrato: {
         frequency: 2.0, // 0 < x ≤ 14
         depth: 0.5      // 0 < x ≤ 1
+    },
+    // Rotates the sound around the stereo channels/user headphones aka Audio Panning. It can produce an effect similar to: https://youtu.be/QB9EB8mTKcc (without the reverb)
+    rotation: {
+        rotationHz: 0 // the best value would be 0.2, It is he frequency the audio should rotate around the listener, in Hertz.
+    },
+    //Distortion effect. It can generate some pretty unique audio effects.
+    distortion: {
+        sinOffset: 0,
+        sinScale: 1,
+        cosOffset: 0,
+        cosScale: 1,
+        tanOffset: 0,
+        tanScale: 1,
+        offset: 0,
+        scale: 1
+    } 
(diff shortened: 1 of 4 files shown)


IMPLEMENTATION.md
```diff
@@ -176,7 +176,25 @@ JSON comments are for illustration purposes only, and will not be accepted by th
     vibrato: {
         frequency: 2.0, // 0 < x ≤ 14
         depth: 0.5      // 0 < x ≤ 1
+    },
+    // Rotates the sound around the stereo channels/user headphones aka Audio Panning. It can produce an effect similar to: https://youtu.be/QB9EB8mTKcc (without the reverb)
+    rotation: {
+        rotationHz: 0 // the best value would be 0.2, It is he frequency the audio should rotate around the listener, in Hertz.
```
freyacodes: What does “best” mean in this context?

```suggestion
        rotationHz: 0 // The best value would be 0.2, It is he frequency the audio should rotate around the listener, in hertz.
```

Mandruyd: In this context "best" means that the default given value there sounds as similar as to the youtube example there, without the reverb of course.

freyacodes: So “best” is totally subjective? That seems misleading

freyacodes:
```suggestion
        rotationHz: 0 // The frequency of the audio rotating around the listener in Hz. 0.2 is similar to the example video above.
```

freyacodes (approved):

Muh9049:
Rotation works for me, however, the distortion filter doesn't work. Unless I'm doing something wrong?

freyacodes:
Please post relevant logs

Muh9049:
image

This is what it logs. However, nothing actually ends up being applied as a filter.

freyacodes:
I believe that is the functional equivalent of setting your volume to 100%

Muh9049:
I've tried a couple of different things with different adjustments for sinOffset, sinScale, etc. but the effect just doesn't get applied and the audio stays the same. Someone else in the JDA server said it doesn't seem to work for them either

Also just to double-check, I also lowered the volume of my music to 5% and then tried using my distortion command, however, nothing happens

PR #436, opened 2021-02-15
Bump lavaplayer to 1.3.71

ToxicMushroom:
regression fixes for 1.3.68+
some mp4 formats threw exceptions in those versions


topi314:
.71 is out with additional fixes https://github.com/sedmelluq/lavaplayer/issues/582

PR #437, opened 2021-02-16
Docker: Allow overriding the default command used by the image

james7132:
Currently, the image does not allow setting custom JVM flags (all instances are bound to use `-Xmx4G`), which makes it impossible to tune Lavalink for low memory environments.

### DIFF
--- a/LavalinkServer/docker/Dockerfile
+++ b/LavalinkServer/docker/Dockerfile
@@ -9,4 +9,4 @@ WORKDIR /opt/Lavalink
 
 COPY Lavalink.jar Lavalink.jar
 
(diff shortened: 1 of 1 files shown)


topi314:
https://docs.docker.com/engine/reference/run/#entrypoint-default-command-to-execute-at-runtime
you can just override the entrypoint

sedmelluq:
You can specify JVM parameters either with `_JAVA_OPTIONS` or `JAVA_TOOL_OPTIONS` environment variables, no need to touch entrypoint or command.

james7132:
As far as I can tell, for docker-compose, and some container orchestration environments, overriding entrypoint requires explicitly specifying it for each container, whereas CMD is overridable via a config file.

I was not aware of the `*_JAVA_OPTIONS` env variable, though it seems like it just appends it to the end of every `java` command sent. In the case multiple values are specified for a memory limit, does it just use the last one in the command?

PR #438, opened 2021-02-16
fix: connected field on playerUpdate is missing

Allvaa:
the connected field isn't set in player.state. so use a declared state variable instead, which the connected field is added there


freyacodes (approved):
Thanks

PR #442, opened 2021-02-23
fix: players are not deleted even using destroy op

Allvaa:
well, the players actually deleted, but when the `gatewayClosed` event emitted, it creates a new one because of `this@SocketContext.getPlayer(guildId)` on the event

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/SocketContext.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/SocketContext.kt
@@ -105,11 +105,12 @@ class SocketContext internal constructor(
     /**
      * Gets or creates a voice connection
      */
-    fun getVoiceConnection(guild: Long): VoiceConnection {
-        var conn = koe.getConnection(guild)
+    fun getVoiceConnection(player: Player): VoiceConnection {
+        val guildId = player.guildId.toLong()
+        var conn = koe.getConnection(guildId)
         if (conn == null) {
-            conn = koe.createConnection(guild)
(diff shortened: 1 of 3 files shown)


LavalinkServer/src/main/java/lavalink/server/io/SocketContext.kt
```diff
@@ -181,23 +182,23 @@ class SocketContext internal constructor(
         koe.close()
     }
 
-    private inner class EventHandler(private val guildId: String) : KoeEventAdapter() {
+    private inner class EventHandler(private val player: Player) : KoeEventAdapter() {
```
Allvaa, line 185: i changed the parameter to `player` so that when sending player updates, it doesn't need to call `this@SocketContext.getPlayer(guildId)` which caused the bug

eepymoonshine:
Thanks for the fix. For a explanation of this bug:

This was introduced on https://github.com/Frederikam/Lavalink/commit/7ef3f3890951655ad143d1739b2c29122b486b2b (which explains why my old instance running a build from early January was fine)

When the player emits the gatewayClosed event (which will happen after the player is deleted), https://github.com/Frederikam/Lavalink/commit/7ef3f3890951655ad143d1739b2c29122b486b2b#diff-69d704c1fea4e5422f61cba6d43b943a69202a8ff81995fb3ebcc74d94f7511eR196 will just create a new Player instance, effectively making the "deletion" of the first player object not happen in practice (even though it just adds a completely new one, so the memory overhead should be lower for the leaky object)

freyacodes (approved):
Nice catch

PR #445, opened 2021-02-26
fix for lavalink not continue playing after voice channel move

topi314:
currently lavalink stops playing when moved to a new voice channel
you need to explicit resume to make it play again. At the same time ll does not send anything to the lavalink-client stating it is paused.
I think resuming would be the best option but I'm open for input from your side.
I'm not sure if you have a better impl for that. Feel free to edit these changes

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
@@ -30,7 +30,9 @@ class WebSocketHandlers(private val contextMap: Map<String, SocketContext>) {
         //discord sometimes send a partial server update missing the endpoint, which can be ignored.
         endpoint ?: return
 
-        context.getVoiceConnection(context.getPlayer(guildId)).connect(VoiceServerInfo(sessionId, endpoint, token))
+        val player = context.getPlayer(guildId)
+        val conn = context.getVoiceConnection(player)
+        conn.connect(VoiceServerInfo(sessionId, endpoint, token)).apply { player.provideTo(conn) }
     }
 
     fun play(context: SocketContext, json: JSONObject) {


LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
```diff
@@ -30,7 +30,9 @@ class WebSocketHandlers(private val contextMap: Map<String, SocketContext>) {
         //discord sometimes send a partial server update missing the endpoint, which can be ignored.
         endpoint ?: return
 
-        context.getVoiceConnection(context.getPlayer(guildId)).connect(VoiceServerInfo(sessionId, endpoint, token))
+        val player = context.getPlayer(guildId)
+        val conn = context.getVoiceConnection(player)
+        conn.connect(VoiceServerInfo(sessionId, endpoint, token)).apply { player.provideTo(conn) }
```
freyacodes, line 35: Why the `apply` function here?

topi314: Idk it worked 🤔
I'm not super into CompleteAbleFutures

topi314: does it even need to wait for that CompleteAbleFuture to be finished?
I tried it without(sounds like a race condition to me then) and it worked but is that the correct way?

freyacodes: You're calling this method: https://kotlinlang.org/docs/scope-functions.html#apply

Moving the contents of the apply block to the next line is functionally equivalent here

freyacodes: The code doesn't look prone to race conditions as the old socket is disconnected and thrown away

topi314: ah alright ty! ❤

PR #460, opened 2021-03-27
Fix distortion filter typo

PredaaA:
The distortion filter was not working (see https://github.com/Frederikam/Lavalink/pull/417#issuecomment-775996087), this PR fixes it.
Thanks `Axel#3456` on JDA server that reported it.


freyacodes:
This typo should be corrected, but how does this PR change the way the code works?

Muh9049:
Distortion wasn't working for me for a while either. I tried this PR locally a few minutes ago and compiled a .jar file and it indeed fixed it and now distortion works again for me.

PredaaA:
So actually I tried to understand why it's not working if we send "distortion" in OP data, but it's just because it's expecting "distorsion" instead, by using the second the filter works.
So yeah basically that PR is only a typo fix, and making it match documentation too.

freyacodes (approved):
I see now that it matters because we're doing JSON deserialization on this class

PR #461, opened 2021-03-30
docu for no selectedTrack

topi314:
`audioPlaylist.getSelectedTrack()` may return `null` which then causes `Integer selectedTrack` to be `-1` as there is no `null` object in `audioPlaylist.getTracks()`

returning `-1` here causes the lavalink-client to throw a nice `Index -1 out of bounds` exception you can see here
https://hastebin.de/cidagilavo.yaml
https://github.com/Frederikam/Lavalink-Client/blob/a216e469638007463bbf641d1ababd7132129625/src/main/java/lavalink/client/io/LavalinkRestClient.java#L248

also there seem to be a problem with lavaplayer & yt playlists again:
https://github.com/sedmelluq/lavaplayer/issues/602


freyacodes:
This is a breaking change. Would probably be better to fix Lavalink-Client and to document the `-1`

topi314:
fair enough
I'll do that alter 👍🏻

topi314:
it can't be null right 🤔

topi314:
should I instead rebase this on master?

freyacodes:
Sure

topi314:
god lemme just open a new pr lmao

PR #478, opened 2021-04-18
source name in track info

topi314:
the track source name is already given in the raw track data and should be useful to also have in the `/loadtracks`, `/decodetrack` & `/decodetracks` endpoints


freyacodes (changes requested):
The documentation should reflect how multiple endpoints are being changed

freyacodes:
(Might want to cherrypick 3c16516)

topi314:
wanna cherrypick https://github.com/freyacodes/Lavalink/pull/480 after being merged :)

freyacodes (approved):

PR #489, opened 2021-05-31
More architectures JDA-NAS

Stonedestroyer:
Adds support for more architectures, will follow up with PR to udp-queue natives by @aikaterna .

PR #490, opened 2021-05-31
JDA-NAS Toggle

Stonedestroyer:
Makes JDA-NAS togglable.

With in server property jdanas false or true. Depends on #489, also optional. And can be added without that PR. Set to true to mimic default behavior. 

Not sure if there's any other place to document this change.

### DIFF
--- a/LavalinkServer/application.yml.example
+++ b/LavalinkServer/application.yml.example
@@ -13,6 +13,7 @@ lavalink:
       mixer: true
       http: true
       local: false
+    jdanas: true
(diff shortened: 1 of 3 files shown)


freyacodes:
Why would someone want to disable [UDP-Queue](https://github.com/KyokoBot/koe/tree/master/ext-udpqueue)?

Note that JDA-NAS is no longer used.

Stonedestroyer:
On a further thought, there may not be much point to disabling it if we control what natives get loaded. Probably would have made more sense if it was no check for this PR but even then probably shouldn't be left to user to decide. Unless otherwise I will close this in a few.

PR #492, opened 2021-05-31
OSHI bump

Stonedestroyer:
Oshi bump, should fix stat related issues and may supersede #211

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/StatsTask.java
+++ b/LavalinkServer/src/main/java/lavalink/server/io/StatsTask.java
@@ -90,7 +90,7 @@ public class StatsTask implements Runnable {
         if(prevTicks == null) {
             prevTicks = hal.getProcessor().getSystemCpuLoadTicks();
         }
-        cpu.put("systemLoad", hal.getProcessor().getSystemCpuLoadBetweenTicks(cpuLoad));
+        cpu.put("systemLoad", hal.getProcessor().getSystemCpuLoadBetweenTicks(prevTicks));
         prevTicks = hal.getProcessor().getSystemCpuLoadTicks();
         double load = getProcessRecentCpuUsage();
         if (!Double.isFinite(load)) load = 0;
--- a/LavalinkServer/src/main/java/lavalink/server/io/StatsTask.java
+++ b/LavalinkServer/src/main/java/lavalink/server/io/StatsTask.java
@@ -42,7 +42,7 @@ public class StatsTask implements Runnable {
 
     private final SystemInfo si = new SystemInfo();
     private final HardwareAbstractionLayer hal = si.getHardware();
-    private long[] cpuLoad;
+    private long[] prevTicks;
 
     StatsTask(SocketContext context, SocketServer socketServer) {
         this.context = context;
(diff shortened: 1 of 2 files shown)


aikaterna:
I have tested this on the machines I have mentioned in the linked PR, a Win2016 vm on an RHEL device, and an x86 Mac both running Adopt OpenJDK 11 builds and the stats issue mentioned is not present. This seems to work as intended.

freyacodes:
This needs to be tested on Zulu as well

aikaterna:
Will do, do you have a preference on what machine(s)? Also any specific java versions?

freyacodes:
At least Java 11 on Linux AMD64 as well as a newer version

aikaterna:
Looks good here too. The test song played through on all attempts with no repeating stats errors or any issues.

Azul Java 11:
https://gist.githubusercontent.com/aikaterna/ccab7e2144511e2de37be4da9b2028ff/raw/9db7d42e7fa4d30a48e0e185f4cce07c3e4a6221/oshi_linux64_java11.txt

Azul Java 13:
https://gist.githubusercontent.com/aikaterna/553a4f82e4c4eb1c8117c2659515b742/raw/81d69085153241d9204cbf7bb8fd9596780046a3/oshi_linux64_java13.txt

Azul Java 15:
https://gist.githubusercontent.com/aikaterna/74a41721d4fc6e8e00a6d2b9d30ff5ae/raw/53b57c45e1a7e74de9854f39619d14aadd826ee5/oshi_linux64_java15.txt

Happy to test on any other combos you would like to see.

LavalinkServer/src/main/java/lavalink/server/io/StatsTask.java
```diff
@@ -82,13 +84,14 @@ private void sendStats() {
...
-
 
         JSONObject cpu = new JSONObject();
         cpu.put("cores", Runtime.getRuntime().availableProcessors());
-        cpu.put("systemLoad", hal.getProcessor().getSystemCpuLoad());
+        if(cpuLoad == null) {
+            cpuLoad = hal.getProcessor().getSystemCpuLoadTicks();
+        }
+        cpu.put("systemLoad", hal.getProcessor().getSystemCpuLoadBetweenTicks(cpuLoad));
+        hal.getProcessor().getSystemCpuLoadTicks();
```
freyacodes: What is this line doing?

Stonedestroyer: Fixed up the logic there, but the idea is

Set prevTicks if not null, continue and add the value returned by getSystemCpuLoadBetweenTicks which is a Double, it checks the ticks between the first value and the second to get a more accurate baseline.

After that I assign the new ticks to prevTicks so we can continue on. I have commented it.

Stonedestroyer: These are relevant <https://github.com/oshi/oshi/blob/596d27a91f05d0393a44909e83ef244c0f654c1f/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java#L108> and <https://github.com/oshi/oshi/blob/596d27a91f05d0393a44909e83ef244c0f654c1f/oshi-core/src/main/java/oshi/hardware/CentralProcessor.java#L136>

LavalinkServer/src/main/java/lavalink/server/io/StatsTask.java
```diff
@@ -41,6 +41,9 @@
     private final SocketServer socketServer;
 
     private final SystemInfo si = new SystemInfo();
+    private final HardwareAbstractionLayer hal = si.getHardware();
+    // CPU ticks used for calculations in CPU load.
```
freyacodes:
```suggestion
    /** CPU ticks used for calculations in CPU load. */
```

PR #546, opened 2021-10-23
Fix: when the player gets moved into another channel the session is r…

Minestroide:
Hi, I would like to do this pull request because when the player gets moved into another channel (i'm currently using Lavalink.py and Discord.py for my bot) the session is resumed but no audio is played in the channel.

Hope it helps.

Regards.

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt
@@ -26,6 +26,10 @@ class WebSocketHandlers(private val contextMap: Map<String, SocketContext>) {
         endpoint ?: return
 
         context.getVoiceConnection(guildId).connect(VoiceServerInfo(sessionId, endpoint, token))
+
+        //fix: when the player gets moved into another channel the session is resumed but no audio is played.
+        val conn = context.getVoiceConnection(guildId)
+        context.getPlayer(guildId.toString()).provideTo(conn)
     }
 
     fun play(context: SocketContext, json: JSONObject) {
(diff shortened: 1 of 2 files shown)


topi314:
I tried this locally and in production
locally it works
in production it wont

its the same as with my changes on dev in feb see: https://github.com/freyacodes/Lavalink/pull/445

Minestroide:
Oh I see, I'm sorry I didn't notice that pull request. In anyway that's strange because I tried to explictly resume the track but that didn't work. Until now I fixed that by just "replaying" the song and letting it start from the point it was stopped but I didn't like that solution and I tried this and it works for me.

Minestroide:
I'll close this since it has already been merged in the dev branch.

topi314:
Don't get me wrong
My "fix" isn't even working

Also the best solution currently is to wait like 0.5/1s and pause & resume

Then it works without problem

Minestroide:
Ok, thanks for your clarifications, don't worry I wasn't getting you wrong, It's just that it's useless if it doesn't work in production. If I'll have the time I'll try to find another solution and then make a new pull request.

topi314:
Would be nice to finally resolve this
👍

Walkyst:
After some testing found that if add artificial delay in around 300-500ms `Thread.sleep(500L)` everything will work fine, so im not completely sure what cause it, but seems to me that this is most likely the reason in koe

The more higher delay then high success rate it have, for example if use 200ms then switching between Russia and Europe regions, sometime Europe is not output audio, only after pause-unpause but if use 300ms then everything okay
```kt
val player = context.getPlayer(guildId)
val conn = context.getVoiceConnection(player)
conn.connect(VoiceServerInfo(sessionId, endpoint, token)).whenComplete { _, _ ->
    Thread.sleep(500L)
    player.provideTo(conn)
}
```

PR #577, opened 2021-12-09
Fix player is not playing after move between vc or changing regions

Walkyst:
Also, does this lines really required here? Because we always expect to have `voiceUpdate` op for connecting to voice

https://github.com/freyacodes/Lavalink/blob/dev/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt#L76-L77

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/SocketContext.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/SocketContext.kt
@@ -52,7 +52,7 @@ class SocketContext internal constructor(
         private var session: WebSocketSession,
         private val socketServer: SocketServer,
         val userId: String,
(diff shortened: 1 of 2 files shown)


Deivu:
> Also, does this lines really required here? Because we always expect to have `voiceUpdate` op for connecting to voice
> 
> https://github.com/freyacodes/Lavalink/blob/dev/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandlers.kt#L76-L77

I removed that line and on my testing, I never had any issues on removing that line so far, personally I think it is not required to be on play method as well

freyacodes:
I believe those lines are necessary because the player can be destroyed between voice server updates

freyacodes (approved):

PR #582, opened 2021-12-19
Bump lavaplayer and add YT config for avoiding age restrictions

Walkyst:
How to get these keys explained here https://github.com/Walkyst/lavaplayer-fork/issues/18


freyacodes (changes requested):
Thanks for your PR!

For consistency these new properties should probably be in the ServerConfig class.

freyacodes (approved):

PR #583, opened 2021-12-19
add onDestroyPlayer event for plugins to use

topi314:
this event can be helpful to get rid of player specific data saved


topi314:
in my case I save which sponsorblock categories a player should retrieve & skip
discarding that data when a player is being destroyed would be helpful

freyacodes (approved):

PR #595, opened 2022-01-06
feat: http proxy config for httpAudioSourceManager

ToxicMushroom:
Useful for stopping bad actors from IP-grabbing your main music server and attacking it.
Tested with squidproxy

### DIFF
--- a/LavalinkServer/application.yml.example
+++ b/LavalinkServer/application.yml.example
@@ -29,8 +29,10 @@ lavalink:
       #PAPISID:      ""
       #PSID:         ""
     #httpConfig:
-      #proxyHost: "localhost"
-      #proxyPort: 3128
+      #proxyHost: "localhost" # Hostname of the proxy, (ip or domain)
+      #proxyPort: 3128 # Proxy port, 3128 is the default for squidProxy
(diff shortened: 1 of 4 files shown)


Walkyst:
How it supposed to connect if proxy behind some basic auth like login and password?

ToxicMushroom:
> How it supposed to connect if proxy behind some basic auth like login and password?

I've added an option for authentication, I don't have a setup to test this though

LavalinkServer/application.yml.example
```diff
@@ -28,6 +28,9 @@ lavalink:
     #youtubeConfig: # Required for avoiding age restrictions by YouTube. Guide https://github.com/Walkyst/lavaplayer-fork/issues/18
       #PAPISID:      ""
       #PSID:         ""
+    #httpConfig:
+      #proxyHost: "localhost"
+      #proxyPort: 3128
```
freyacodes: This needs to be documented with comments, and this is missing username and password

freyacodes (changes requested):

freyacodes:
It still needs a description of why you would ever want to set up a proxy

freyacodes (approved):

Issue #599, opened 2022-01-06
Required library was not found (dev branch/builds)


Muh9049:
## Description
For some reason using the official dev builds to apply the timescale filters returns this error and completely breaks the track: <https://srcb.in/YHbQc8HPlS>

Upon further digging into this issue, it seems that using the official master builds does not cause the track to crash and give that error and instead applies the timescale filter like it should even though it’s using Walkyst’s fork just like dev branch. Does anyone know why this might be the case and a workaround/fix?

master builds: https://ci.fredboat.com/viewType.html?buildTypeId=Lavalink_Build&tab=buildTypeStatusDiv&guest=%21&branch_Lavalink=%3Cdefault%3E
dev builds: https://ci.fredboat.com/viewType.html?buildTypeId=Lavalink_Build&branch_Lavalink=refs%2Fheads%2Fdev&tab=buildTypeStatusDiv&guest=!

**Note**: This issue happens on both Windows and Linux.

## Version info
Output of `java -version` (windows):
```
openjdk 13.0.2 2020-01-14
OpenJDK Runtime Environment (build 13.0.2+8)
OpenJDK 64-Bit Server VM (build 13.0.2+8, mixed mode, sharing)
```

Output of `java -version` (linux):
```
openjdk 16.0.1 2021-04-20
OpenJDK Runtime Environment (build 16.0.1+9-Ubuntu-120.04)
OpenJDK 64-Bit Server VM (build 16.0.1+9-Ubuntu-120.04, mixed mode, sharing)
```

szymonwilczek:
Try using this version of Lavalink: https://github.com/Cog-Creators/Lavalink-Jars/releases, and let me know

Muh9049:
> Try using this version of Lavalink: https://github.com/Cog-Creators/Lavalink-Jars/releases, and let me know

At least from what I've heard, I'm sure it would but that doesn't solve the initial issue in the actual repo though. I'd rather use the official builds or build my own jars but I can't use it fully if I get the library not found error when applying timescale filters.

The problem is that the error is only occuring with the dev builds and not the master builds even though it uses the same LP fork, ladaDsp version, etc. Since that's the case, I assume something must've gotten messed up in a recent dev branch commit.
Keep in mind the dev builds never gave me this sort of issue until recently.

Muh9049:
[...]

PR #600, opened 2022-01-06
fix: missing timescale natives

Muh9049:
Fixes the error when trying to apply timescale filters using the dev branch: #599


freyacodes (approved):
Thanks!

PR #605, opened 2022-01-13
Format client libraries as a table

devoxin:
First time really working with markdown tables so please let me know if there's anything that can be improved!

I've restructured the list of client libraries as a table as this should hopefully improve readability and define a proper format for future client library additions. I've also removed any broken links and extremely outdated libraries, except those that I believe could still work with current Lavalink versions, as appears to be the case with a handful of JavaScript libraries for example.

I've also marked any libraries that have not received commits on their respective repositories in more than a year, as unmaintained. These have not been removed as there is a good chance that they may still work with current Lavalink versions (as the API version hasn't changed in the last year), however their feature coverage may be lacking (filter support, for example).

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,39 +69,40 @@ While major, minor and patch will do an optimum effort to adhere to [Semantic Ve
 
 
 ## Client libraries:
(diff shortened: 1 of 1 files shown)


freyacodes (approved):

freyacodes:
Looks great!

Issue #569, opened 2021-11-20
Tracks with a specified `endTime` don't go to the next in queue as expected


zedseven:
## Description
When a track with a specified end time reaches the time specified, it is stopped rather than being properly finished. This means it has to be "skipped" upon finishing, which requires additional overhead and doesn't seem to make sense.

When I say "properly finished", I specifically mean that I expect the track to end and the next in queue to play exactly as if the track had played out completely. I feel this is the way most users would expect it to behave.

If I'm somehow mistaken or this behaviour is working as designed, I'd love to know why. Thank you for your work on this library and tool.

## Version info
**Client used:** [lavalink-rs (with the original issue I raised for this)](https://gitlab.com/vicky5124/lavalink-rs/-/issues/5)

Output of `java -version`:
```
openjdk version "16" 2021-03-16
OpenJDK Runtime Environment AdoptOpenJDK (build 16+36)
OpenJDK 64-Bit Server VM AdoptOpenJDK (build 16+36, mixed mode, sharing)
```

My log:
[spring.log](https://github.com/freyacodes/Lavalink/files/7573782/spring.log)

topi314:
this has been fixed in https://github.com/freyacodes/Lavalink/pull/609 and `3.5rc1`

PR #609, opened 2022-01-14
Correctly emit 'FINISHED' in TrackEndEvent for TrackEndMarkers

devoxin:
This PR introduces an `endMarkerHit` boolean within `linkPlayer` which is set to `true` by the `TrackEndMarkerHandler` class when the provided `endTime` is hit (if applicable).

This boolean is then used by the `EventEmitter` class to establish whether to emit the *true* end reason, or `FINISHED`. This is necessary as most libraries use the `reason` in the `TrackEndEvent` to determine whether or not they should play the next song in queue, to avoid unintended behaviour such as double-skipping or playing when a `stop` command has been called, for example.

This seemed like the cleanest way to implement this solution for **#569**, as this should work for all tracks including livestreams.
I wasn't sure whether to implement a custom reason (such as `END_MARKER_HIT`) so libraries could optionally implement special handling or user-facing messages so I opted for the safe route of emitting `FINISHED`, though I can change this upon request.

As always, feedback is welcomed! :)

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/player/EventEmitter.java
+++ b/LavalinkServer/src/main/java/lavalink/server/player/EventEmitter.java
@@ -75,7 +75,12 @@ public class EventEmitter extends AudioEventAdapter {
             out.put("track", JSONObject.NULL);
         }
 
-        out.put("reason", endReason.toString());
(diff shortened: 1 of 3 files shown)


freyacodes (approved):
Not a pretty implementation but I suppose this will work

PR #610, opened 2022-01-20
Add wavelink back to README.md

EvieePy:
Adding WaveLink back to README as it is now un archived and beiing maintained.


devoxin:
Something small - might I suggest grouping it with Pomice (and lavaplayer-py with lavasnek_rs) so the information column lines up? 😅

EvieePy:
I think this is what you mean :)

devoxin:
That looks perfect 🙂

freyacodes (commented):
Are you sure you want to link the 1.0 branch? Maybe it is more helpful for your users if you change the default branch instead

EvieePy:
I did that because we just had one more thing to implement before changing branches. But sure this can wait till that has been merged and I will update accordingly.

freyacodes:
Just make sure you post a new PR

PR #618, opened 2022-02-09
Allow plugins to modify the response of /loadtracks 

viztea:
Added the ability for plugins to modify the `playlistInfo` and `tracks[*].info` objects returned by /loadtracks.

- adds an interface to the plugin api: `AudioLoaderJsonModifier`.
- converts `AudioLoaderRestHandler` and `LoadResult` to kotlin so that I could keep my sanity.
- adds `lavalink.server.allowTrackModifications` to prevent clients from breaking (defaults to false).


freyacodes:
This looks like an interesting addition. I'm currently very busy, so don't expect a proper review before the 25th of February

IMPLEMENTATION.md
```diff
@@ -419,6 +419,8 @@ If the identifier leads to a playlist, `playlistInfo` will contain two propertie
 }
 ```
 
+If enabled by the `lavalink.server.allowTrackModifications` config value, any installed plugins may modify the contents of `playlistInfo` and each `info` object for a track
```
freyacodes, line 422: I think I see where you are coming from but I don't like this. Plugins should be allowed to break the API if they so wish.

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
```diff
@@ -0,0 +1,175 @@
+/*
```
freyacodes, line 1: It's difficult to review this code because you converted the file to Kotlin while in the same commit you implemented the feature

plugin-api/src/main/java/dev/arbjerg/lavalink/api/AudioLoaderJsonModifier.java
```diff
@@ -0,0 +1,28 @@
+package dev.arbjerg.lavalink.api;
+
+import com.sedmelluq.discord.lavaplayer.track.AudioPlaylist;
+import com.sedmelluq.discord.lavaplayer.track.AudioTrack;
+import org.json.JSONObject;
+
+/**
+ * Allows modification of JSON objects returned by the `/loadtracks` endpoint.
+ */
```
freyacodes, line 9: This should document how to use it with a bean

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
```diff
@@ -0,0 +1,175 @@
...
+    companion object {
+        private val log = LoggerFactory.getLogger(AudioLoaderRestHandler::class.java)
+
+        private inline fun <reified T> Any?.intoOrNull(): T? = this as? T
+
+        private fun log(request: HttpServletRequest) = log.info("GET ${request.servletPath}")
+
+        private fun attemptModifier(modifier: AudioLoaderJsonModifier, json: JSONObject, block: AudioLoaderJsonModifier.(JSONObject) -> JSONObject): JSONObject? {
+            return json.clone()
+                .runCatching { modifier.block(this) }
```
freyacodes, line 54: This looks like it will swallow the error without logging it. A simple try/catch would probably be a better and simpler alternative to this method

freyacodes (changes requested):
It looks like `AudioLoaderJsonModifier` is invoked when it's either a bean or when the audio source manager is an implementation of it. This seems unnecessarily convoluted:

* It means that the same modifier can be invoked twice for the same object, especially as audio source managers are often provided as beans.
* It doesn't add much value as modifiers can already be made to only target specific types of managers or track types
* It's confusing, and it's not even documented

viztea:
> It means that the same modifier can be invoked twice for the same object, especially as audio source managers are often provided as beans.

> It doesn't add much value as modifiers can already be made to only target specific types of managers or track types

True, I forget the reason I did this but it's definitely an overlook on my part.

> It's confusing, and it's not even documented

As in not documented in the `PLUGINS.md`? I could write docs for it.

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
viztea: Yeah sorry :/

freyacodes:
> As in not documented in the PLUGINS.md? I could write docs for it.

That would be nice as well but I was more thinking about the documentation in the interface

PR #631, opened 2022-03-08
Allow adding new MediaContainerProbes via plugins

TheEssem:
Just a small edit to `AudioPlayerConfiguration` that adds the ability to add new `MediaContainerProbe` instances. This is useful for adding support for new HTTP/local file formats or adding a new source that depends on some sort of weird codec.

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/config/AudioPlayerConfiguration.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/config/AudioPlayerConfiguration.kt
@@ -1,5 +1,7 @@
 package lavalink.server.config
 
+import com.sedmelluq.discord.lavaplayer.container.MediaContainerProbe
+import com.sedmelluq.discord.lavaplayer.container.MediaContainerRegistry
 import com.sedmelluq.discord.lavaplayer.player.AudioPlayerManager
 import com.sedmelluq.discord.lavaplayer.player.DefaultAudioPlayerManager
 import com.sedmelluq.discord.lavaplayer.source.AudioSourceManager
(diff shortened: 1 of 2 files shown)


LavalinkServer/src/main/java/lavalink/server/config/AudioPlayerConfiguration.kt
```diff
@@ -109,7 +114,7 @@ class AudioPlayerConfiguration {
         if (sources.isTwitch) audioPlayerManager.registerSourceManager(TwitchStreamAudioSourceManager())
         if (sources.isVimeo) audioPlayerManager.registerSourceManager(VimeoAudioSourceManager())
         if (sources.isMixer) audioPlayerManager.registerSourceManager(BeamAudioSourceManager())
-        if (sources.isLocal) audioPlayerManager.registerSourceManager(LocalAudioSourceManager())
+        if (sources.isLocal) audioPlayerManager.registerSourceManager(LocalAudioSourceManager(mcr))
```
freyacodes, line 117: Is this additive? Are the default probes removed if the collection is empty?

viztea: > Is this additive?

Nope

TheEssem: No, LocalAudioSourceManager when initialized without any arguments uses the default probe list and `MediaContainerRegistry#extended()` simply adds onto the default probe list.

freyacodes: Ah yes, I see

freyacodes (changes requested):
This feature needs to be documented in PLUGINS.md otherwise plugin developers have no chance of knowing about it

TheEssem:
Sorry about that, not sure how it closed itself.

freyacodes (approved):
Thanks! Sorry for the wait

PR #633, opened 2022-03-12
Added additional audio filter

rohank05:
- Added echo audio filter
- Added lavadsp-extended dependencies source: https://github.com/rohank05/lavadsp-extended
- Edited IMPLEMENTATION.md to add echo example

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -226,6 +226,12 @@ Note that filters may take a moment to apply.
     "lowPass": {
         "smoothing": 20.0
     }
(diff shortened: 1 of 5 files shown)


someone508:
Maybe you could create your own lavalink plugin using the lavalink plugin-api with your custom filters? That way the lavalink doesn't need an extra dependency which may need regular updates (if you plan to add more filters idk), and if someone doesn't want more filters, just don't use your plugin.

rohank05:
Ahh i didn't realise

PR #641, opened 2022-03-31
Update spring dependencies

Fabricio20:
This commit should tackle any possibility of CVE-2022-22965 being exploited on LavaLink. As far as the current information available states, LavaLink is not vulnerable as we are using the Undertow container and running as a deployable jar instead of a war-file.

More information is available on spring's blog: [https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement](https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement).

This dependency update required an upgrade to Gradle 6.9 as the spring-boot dependency now requires at least gradle 6.8, since 7.0 made changes to how a lot of APIs work, I've sticked with 6.9 for now.

### DIFF
--- a/LavalinkServer/build.gradle
+++ b/LavalinkServer/build.gradle
@@ -12,6 +12,7 @@ apply plugin: "com.adarshr.test-logger"
 description = 'Play audio to discord voice channels'
 mainClassName = "lavalink.server.Launcher"
 version = "${versionFromTag()}".toString()
+
 ext {
     moduleName = 'Lavalink-Server'
 }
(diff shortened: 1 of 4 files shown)


freyacodes:
As you said, Lavalink should be unaffected by this vulnerability but I agree that we should update as a precaution.

I can't accept this change to the master branch directly, so I have changed the base. After merging I can look into releasing a new version. Please do me a favour and fix the conflict.

Fabricio20:
I think that's done, I also removed the `doLast` step from the `bootJar` configuration since it seems to do the same thing as `archiveFileName` does. Let me know if you need anything else changed.

freyacodes (approved):

PR #654, opened 2022-05-01
fix(typo): "JDAA" should be "JDA"

rilysh:


freyacodes:
Actually not a typo. JDAA refers to the long-obsolete JDA-Audio

PR #657, opened 2022-05-21
Document search prefixes in IMPLEMENTATION

devoxin:


freyacodes (approved):
I felt the need to change the wording, but it's good that we finally document this

PR #665, opened 2022-06-05
Add log message when application is ready

Jackenmen:
This new log message marks the moment the application is ready more clearly now that there are 2 'Started Launcher' log messages (a result of there now being two spring applications - PluginManager and LavalinkApplication).

Additionally, this is somewhat helpful for people running LL through subprocesses that want to programmatically detect when the application starts to accept connections [such as we do over at Red](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/audio/manager.py) as it's probably more reliable over checking that "Started Launcher" message was logged twice which is probably somewhat more likely to change in future as it depends on spring init.

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/Launcher.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/Launcher.kt
@@ -34,6 +34,7 @@ import org.springframework.boot.autoconfigure.SpringBootApplication
 import org.springframework.boot.builder.SpringApplicationBuilder
 import org.springframework.boot.context.event.ApplicationEnvironmentPreparedEvent
 import org.springframework.boot.context.event.ApplicationFailedEvent
+import org.springframework.boot.context.event.ApplicationReadyEvent
 import org.springframework.context.ApplicationListener
 import org.springframework.context.ConfigurableApplicationContext
 import org.springframework.context.annotation.ComponentScan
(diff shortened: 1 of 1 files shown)


topi314:
Just out of curiosity 
why don't you just start to connect to lavalink and retry until you connect instead?

looking at the logs and starting the connection process then seems more complicated tbh

Jackenmen:
Reasons that come to my mind:
- avoiding spurious logs about failed attempts
  - this is especially a problem for low-end devices such as Raspberry Pi 3 where Lavalink may take up to a minute to connect
    - even running things on a VM with a single-core assigned to it can slow startup a bit
  - either way, there are going to be some failed attempts that would always be logged and just concern people unnecessarily (and this can translate to people reporting these "issues")
- connecting to a completely different Lavalink server than the managed one if there is already one running on the computer
  - if we're spawning a managed Lavalink process, we don't want to connect to something else; there's an explicit configuration option for the user to be able to run Lavalink externally - if they're not using it, they probably don't want to use some unmanaged Lavalink server (and it's possible they misconfigured something which is easier for them to detect if they see a failure)

Overall, I also wouldn't say that looking through console output particularly complicates the process management as it may initially seem.

freyacodes (approved):
The approach I'd take would have been to just test if the port is open with a simple TCP socket without sending data. Spring Boot doesn't log errors for that. Nevertheless I don't mind merging this.

freyacodes:
Note that I've replaced Application to Lavalink

PR #672, opened 2022-06-20
Switch to kotlin gradle dsl and use version catalogs

viztea:
* switches from groovy to kotlin for most gradle files
* uses version catalogs
* updates gradle version to latest


viztea:
~~I changed some annotation imports due to them not being found anymore, may have been an issue with my IDE, not sure. I'll see if I can revert it and still be able to compile~~

melike2d:
fuck

LavalinkServer/build.gradle.kts
```diff
@@ -0,0 +1,144 @@
...
+
+tasks {
+    build {
+        doLast {
+            println("Version: $version")
+        }
+    }
+
+    processResources {
+        val buildNumber = if (System.getenv("CI") != null) {
```
devoxin, line 74: Could this not be changed to something like
```kt
System.getenv("CI")?.let { System.getenv("BUILD_NUMBER") ?: "Unknown" } ?: "Unofficial"
```

freyacodes: I think the original proposal is more readable

freyacodes (approved):
I was not aware of the version catalog feature. That's a welcome addition

PR #689, opened 2022-07-12
add github actions to build jars & docker images

topi314:
to whitelist github actions:
```
actions/checkout@v3
actions/setup-java@v3
gradle/gradle-build-action@v2
actions/upload-artifact@v3
actions/download-artifact@v2
docker/metadata-action@v4
docker/setup-qemu-action@v2
docker/setup-buildx-action@v2
docker/build-push-action@v3
softprops/action-gh-release@v1
```

to configure secrets:
```
DOCKERHUB_USERNAME
DOCKERHUB_TOKEN
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,98 @@
+name: Build
+
+on:
+  push:
+    paths-ignore:
+      - '**.md'
+  pull_request:
+
+jobs:
+  gradle:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: zulu
+          java-version: 13
+          cache: gradle
+
+      - name: Setup Gradle
+        uses: gradle/gradle-build-action@v2
+
+      - name: Execute Gradle build
+        run: ./gradlew build
+
+      - name: Upload Artifacts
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,99 @@
...
+      - name: Execute Gradle build
+        run: ./gradlew build
+
+      - name: Upload Artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: Lavalink.jar
+          path: LavalinkServer/build/libs/Lavalink.jar
+
+  docker:
```
freyacodes, line 36: Could you make this skip if there is no credentials available? It seems to be failing instead

.github/workflows/build.yml
```diff
@@ -0,0 +1,99 @@
+name: Build
+
+on:
+  push:
+    paths-ignore:
+      - '**.md'
+  pull_request:
```
freyacodes, line 7: I think this line is unnecessary. Pull requests trigger the push event, making it redundant in this case. By removing this you can simplify the run conditions and remove many of the "skipped" commit statuses

freyacodes (changes requested):

freyacodes (approved):
I'm actually going to merge this now such that we can actually benefit from it. Please continue development in a new PR

.github/workflows/build.yml
topi314: resolved in https://github.com/freyacodes/Lavalink/pull/698

.github/workflows/build.yml
topi314: resolved in https://github.com/freyacodes/Lavalink/pull/698

PR #697, opened 2022-07-26
feat: configureable voice gateway version

KagChi:


devoxin:
This doesn't seem like such a good idea from the perspective that if the voice gateway receives a breaking change in how it operates, the library might not be able to handle the change and could lead to confusion for the end user if Lavalink stops working correctly, or as expected.

PR #698, opened 2022-07-28
fix github actions issues

topi314:
remove on pull_request & block job from running if no docker secrets are provided


freyacodes (changes requested):
There's a syntax error

> The workflow is not valid. .github/workflows/build.yml (Line: 37, Col: 9): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.DOCKERHUB_USERNAME

topi314:
secrets don't seem to be supported in the job ifs uh

PR #742, opened 2022-09-16
feat(plugin-api): userId & clientName getters

viztea:
Adds two methods for getting the name & user id of a client


viztea:
uh, i meant dev smh

PR #754, opened 2022-09-26
removed last mention of build number

topi314:


devoxin:
Looks fine, but any particular reason for removing it?

topi314:
build number was used in the old ci but has been removed with moving to github actions
just forgot to remove that part in there

freyacodes (approved):

PR #761, opened 2022-10-06
chore: update kotlin version

viztea:
Updates the kotlin version to **1.7.20** which is the latest as of me creating this PR, I've also added a note to the plugin implementation docs because this is what initially brought me to create this PR


freyacodes:
This seems like a potentially breaking change for plugins. Perhaps we should hold off with this until we are doing a major release

viztea:
> This seems like a potentially breaking change for plugins. Perhaps we should hold off with this until we are doing a major release

Fair enough

PR #763, opened 2022-10-09
Remove erela.js from client list

viztea:
It has been archived


devoxin:
If its archived, note it in the additional notes column. We can sweep the list when v4 drops

Edit: have just checked out the repo, I suppose it can be removed then actually

devoxin (approved):

PR #770, opened 2022-10-28
update github actions

topi314:
We are waiting on https://github.com/softprops/action-gh-release to update or we switch to a different action


topi314:
https://github.com/ncipollo/release-action
this one seems to be popular but all we need is uploading a jar so...

Lulalaby:
image
might be worth looking in official forks from the [original release action](https://github.com/actions/create-release)

PR #779, opened 2022-11-22
config option to disable specific filters

### LINKED ISSUE #771: [Feature Request] Disable filters via config file
Currently, hosts who don't wish to have filters enabled must create an HTTP blackhole with Nginx, etc, or some other method to redirect requests which would otherwise apply a filter to the player.

Filters can significantly increase the resource usage of a player, as such it would be great if there was a native and supported method to do this (and ideally report that this is done so libs can know ahead of time instead of assuming a filter was applied when it isn't (`/v3/info`) 

Something like the following
```yaml
lavalink:
  server:
    filters: false
```
alternatively
```yaml
lavalink:
  server:
    filters: 
      volume: false 
      timescale: false 
      Vibrato: true
...
```

topi314:
This adds an option to disable individual filters.
All ilters are enabled by default regardless if they are listed in there

https://github.com/freyacodes/Lavalink/issues/771

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandler.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandler.kt
@@ -132,7 +132,7 @@ class WebSocketHandler(
                 "'filters' op. Please switch to use that one, as this op will get removed in v4.")
         loggedEqualizerDeprecationWarning = true
 
-        if (!filterConfig["equalizer"]!!) return log.warn("Equalizer is disabled in the config, ignoring equalizer op")
+        if (filterConfig["equalizer"] == false) return log.warn("Equalizer is disabled in the config, ignoring equalizer op")
 
         val player = context.getPlayer(json.getString("guildId"))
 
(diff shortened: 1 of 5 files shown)


devoxin:
This seems a little redundant. What's the purpose of disabling filters when clients can opt not to send configuration values for a given filter?

LavalinkServer/src/main/java/lavalink/server/io/WebSocketHandler.kt
```diff
@@ -131,6 +132,8 @@ class WebSocketHandler(
                 "'filters' op. Please switch to use that one, as this op will get removed in v4.")
         loggedEqualizerDeprecationWarning = true
 
+        if (!filterConfig["equalizer"]!!) return log.warn("Equalizer is disabled in the config, ignoring equalizer op")
```
devoxin:
```suggestion
        if (filterConfig["equalizer"] == false) return log.warn("Equalizer is disabled in the config, ignoring equalizer op")
```

devoxin (changes requested):

topi314:
See https://github.com/freyacodes/Lavalink/issues/771

devoxin (approved):

PR #780, opened 2022-11-22
option to diable jda nas

### LINKED ISSUE #52: Missing option to disable native audio
When testing LavaLink on new systems it would be nice to have a config option to disable native audio sending, since the internal checks do not account for new kernels/systems.

LavaLink seems to default NAS to enabled, making it impossible to test the resources on (currently) unsupported systems like linux-aarch64.

topi314:
this adds an option to disable jda nas

### DIFF
--- a/LavalinkServer/application.yml.example
+++ b/LavalinkServer/application.yml.example
@@ -12,6 +12,7 @@ lavalink:
       vimeo: true
       http: true
       local: false
+    jdaNASDisabled: false # Disables JDA Native Audio System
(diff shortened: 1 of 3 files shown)


LavalinkServer/application.yml.example
```diff
@@ -12,6 +12,7 @@ lavalink:
       vimeo: true
       http: true
       local: false
+    jdaNASDisabled: false # Disables JDA Native Audio System
```
devoxin: To avoid the configuration file becoming polluted, perhaps we can allow for a bufferDuration of 0ms to implicitly disable NAS. Documented, of course.

topi314: Yeah that would work too
I remember us actually talking about this well

devoxin (approved):

PR #781, opened 2022-11-22
autocreate default config if none was found

### LINKED ISSUE #168: Feature request: Auto create default configuration.
Auto create default configuration file if does not exists in lavalink directory, instead of download manually configuration and put into directory.

topi314:


topi314:
the only issue here is that the logging is a bit early and is not styled due to that
image

I'd like to generate the config before initializing all components tho

LavalinkServer/src/main/java/lavalink/server/Launcher.kt
```diff
@@ -120,6 +121,23 @@ object Launcher {
             println(getVersionInfo(indentation = "", vanity = false))
             return
         }
+
+        val config = File("./application.yml")
+        if (!config.exists()) {
+            log.info("No application.yml found, creating one...")
+            Launcher::class.java.getResource("/application.yml.example")
```
devoxin: Perhaps we should check if this is null and log an error just in case.

LavalinkServer/src/main/java/lavalink/server/Launcher.kt
```diff
@@ -136,6 +135,9 @@ object Launcher {
                         it.copyTo(out)
                     }
                 }
+            } ?: run {
```
devoxin: `run` shouldn't be necessary here, and I'm not sure if this would run if `openStream` returns null.

```kotlin
Launcher::class.java.getResource("/application.yml.example")
    ?.openStream()
    ?.use {
        if (!config.createNewFile()) {
            log.error("Unable to create application.yml")
            return
        }
        config.outputStream().use { out ->
            it.copyTo(out)
        }
    }
    ?: log.error("Unable to find application.yml.example in resources")
```

topi314: `openStream` shouldn't return `null` I just realized

devoxin (approved):

PR #782, opened 2022-11-24
Add Coglink to client libraries

ThePedroo:


freyacodes (approved):
A C client? That's neat

PR #821, opened 2023-01-28
Adding the custom version of Erela.js!

CustomTag:
I have implemented a custom version of Erela.js created by Tomato6966. It supports both the REST API and WebSocket, even for users using an older version of Lavalink. The custom version of Erela.js supports both v3 and v4 of the REST API, allowing users to choose which version they prefer, and even v2 if necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,10 +80,10 @@ Additional labels for release candidates are available as extensions to the `MAJ
 | [Lavacord](https://github.com/lavacord/lavacord)                                                      | Node.js  | **Any**                 | ❌                |                                 |
 | [FastLink](https://github.com/ThePedroo/FastLink)                                                     | Node.js  | **Any**                 | ❌                |                                 |
 | [Moonlink.js](https://github.com/1Lucas1apk/moonlink.js)                                              | Node.js  | **Any**                 | ❌                |
-| [@skyra/audio](https://github.com/skyra-project/audio)                                                | Node.js  | discord.js              | ❌                |Archived                        |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -80,9 +80,10 @@ Additional labels for release candidates are available as extensions to the `MAJ
 | [Lavacord](https://github.com/lavacord/lavacord)                                                      | Node.js  | **Any**                 | ❌                |                                 |
 | [FastLink](https://github.com/ThePedroo/FastLink)                                                     | Node.js  | **Any**                 | ❌                |                                 |
 | [Moonlink.js](https://github.com/1Lucas1apk/moonlink.js)                                              | Node.js  | **Any**                 | ❌                |
-| [@skyra/audio](https://github.com/skyra-project/audio)                                                | Node.js  | discord.js              | ❌                | Archived                        |
+| [@skyra/audio](https://github.com/skyra-project/audio)                                                | Node.js  | discord.js              | ❌                |Archived                        |
```
ooliver1: Why was this change made? It is all misaligned now.

CustomTag: No its not misaligned it just looks like that in the code but not in the Preview

ooliver1: That's the point, the table should still work, but it is an unnecessary change - as it is misaligned.

topi314: if you want this merged please fix the alignment, it should also look good in code

topi314 (changes requested):

README.md
CustomTag: > if you want this merged please fix the alignment, it should also look good in code

I fixed the alignment so you can push the request when ever you feel for it.

PR #843, opened 2023-03-05
Change version of lavaplayer from 1.3.99.2 to 1.4.0

kanishkdebnath-bst:
The new version of lavaplayer adds the fix to twitch stream issue.


topi314:
that change is already on dev and waiting to be released. ref: https://github.com/freyacodes/Lavalink/pull/842

PR #846, opened 2023-03-06
Use kotlinx.serialization

DRSchlaubi:
This PR attempts to make the `protocol` module multiplatform by using [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization) instead of jackson. these are the highlights:
- Make all classes @Serializable
- Use the Kotlin Multiplatform plugin instead of JVM for protocol
  - Don't apply the java plugin to all modules
- Bump Kotlin to 1.8.10 (we need this for generic value classes)
- Move all Lavaplayer related code into server module
- Move v3 classes to `jvmMain`
- Add KotlinxSerialization2HttpMessageConverter and add it to spring configuration


protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/messages.kt
```diff
@@ -0,0 +1,240 @@
...
+        val sessionId: String,
+    ) : Message {
+        constructor(
+            resumed: Boolean,
+            sessionId: String
+        ) : this(Op.Ready, resumed, sessionId)
+
+    }
+
+    @SerialName("playerEvent")
```
viztea: Did you mean `playerUpdate`?

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/player.kt
```diff
@@ -0,0 +1,75 @@
...
+    val artworkUrl: String?,
+    val isrc: String?
+)
+
+@Serializable
+data class VoiceState(
+    val token: String = "",
+    val endpoint: String = "",
+    val sessionId: String = "",
+    val connected: Boolean = false,
```
viztea: `connected` and `ping` were removed.

viztea (changes requested):
brief scroll through

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/messages.kt
DRSchlaubi: Good catch, however ws payloads are the only untested thing rn anyw

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/player.kt
DRSchlaubi: Got it from IMPLEMENTATION.md, guess that's not up to date, will remove

settings.gradle.kts
```diff
@@ -78,6 +78,10 @@ fun VersionCatalogBuilder.common() {
     library("oshi",           "com.github.oshi",      "oshi-core").version("5.7.4")
     library("json",           "org.json",             "json").version("20180813")
 
+    library("kotlinx-serialization-json", "org.jetbrains.kotlinx", "kotlinx-serialization-json").version("1.5.0")
+    library("kotlinx-datetime", "org.jetbrains.kotlinx", "kotlinx-datetime").version("0.4.0")
+    library("ktor-http", "io.ktor", "ktor-http").version("2.2.4")
```
topi314: I don't think ktor is used anywhere rn?

protocol/src/jvmMain/kotlin/dev/arbjerg/lavalink/protocol/v4/util.kt
```diff
@@ -1,21 +1,37 @@
...
+}
+
+fun Exception.Companion.fromFriendlyException(e: FriendlyException) = Exception(
+    e.message,
+    Exception.Severity.fromFriendlyException(e.severity),
+    e.toString()
+)
+
+fun LoadResult.Companion.loadFailed(exception: FriendlyException) =
+    loadFailed(Exception.fromFriendlyException(exception))
```
topi314: let's move this into lavalink-server to remove the lavaplayer dependency

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/stats.kt
```diff
@@ -0,0 +1,79 @@
...
+ * Representation of node stats.
+ *
+ * @property frameStats the frame stats of the node. null if the node has no players or when retrieved via `/v4/stats`
+ * @property players the amount of players connected to the node
+ * @property playingPlayers the amount of players playing a track
+ * @property uptime the uptime of the node in milliseconds
+ * @property memory the memory stats of the node
+ * @property cpu the cpu stats of the node
+ */
+interface Stats {
```
topi314, line 15: is there any reason for this being an interface?

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/filters.kt
```diff
@@ -1,28 +1,21 @@
...
+    val equalizer: Omissible<List<Band>> = Omissible.Omitted(),
+    val karaoke: Omissible<Karaoke?> = Omissible.Omitted(),
+    val timescale: Omissible<Timescale?> = Omissible.Omitted(),
+    val tremolo: Omissible<Tremolo?> = Omissible.Omitted(),
+    val vibrato: Omissible<Vibrato?> = Omissible.Omitted(),
+    val distortion: Omissible<Distortion?> = Omissible.Omitted(),
+    val rotation: Omissible<Rotation?> = Omissible.Omitted(),
+    val channelMix: Omissible<ChannelMix?> = Omissible.Omitted(),
+    val lowPass: Omissible<LowPass?> = Omissible.Omitted(),
+    val pluginFilters: Map<String, JsonElement> = mutableMapOf()
```
topi314, line 18: as discussed with @DRSchlaubi this change might not be avoidable

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/stats.kt
DRSchlaubi: There is no @JsonUnwrapped so the StatsEvent implements this as well

settings.gradle.kts
DRSchlaubi: good catch

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/filters.kt
DRSchlaubi: It would be possible by making a custom serializer but imo it's not worth it, doing it like this also makes it easier for clients to implement it

LavalinkServer/src/main/java/lavalink/server/io/SocketContext.kt
```diff
@@ -253,8 +260,19 @@ class SocketContext(
 
     private inner class WsEventHandler(private val player: LavalinkPlayer) : KoeEventAdapter() {
         override fun gatewayClosed(code: Int, reason: String?, byRemote: Boolean) {
-            val event = Message.WebSocketClosedEvent(code, reason ?: "", byRemote, player.guildId.toString())
-            sendMessage(event)
+            if (version == 3) {
+                val event = Message.WebSocketClosedEvent(code, reason ?: "", byRemote, player.guildId.toString())
+                sendMessage(event)
+            } else {
+                val event = dev.arbjerg.lavalink.protocol.v4.Message.EmittedEvent.WebSocketClosedEvent(
```
topi314: should prob use an import here
I'd say we import the v3 stuff with an alias
not the v4 stuff

DRSchlaubi: we need to import sendV4Message with an alias as the "sendV3Message" is a member function of ISocketContext, hwoever I aggree we should use an alias for v3 event classes

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/loadResult.kt
```diff
@@ -0,0 +1,191 @@
...
+
+    @Serializable
+    data class PlaylistLoaded private constructor(
+        override val loadType: ResultStatus,
+        override val data: Data
+    ) : LoadResult {
+        constructor(data: Data) : this(ResultStatus.PLAYLIST, data)
+
+        @Serializable
+        data class Data(
```
topi314: shouldnt this use `Playlist` here?

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/loadResult.kt
```diff
@@ -0,0 +1,191 @@
...
+
+    @Serializable
+    data class TrackLoaded private constructor(
+        override val loadType: ResultStatus,
+        override val data: Data
+    ) : LoadResult {
+        constructor(data: Data) : this(ResultStatus.TRACK, data)
+
+        @Serializable
+        data class Data(
```
topi314: shouldnt this use `Track` here?

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/loadResult.kt
DRSchlaubi: We can, I just did it so it is more consistent with other types

topi314: I think it would make sense so people can pass this object around in their code without issue

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/loadResult.kt
```diff
@@ -0,0 +1,191 @@
...
+            val info: PlaylistInfo,
+            val pluginInfo: JsonObject?,
+            override val tracks: List<Track>
+        ) : LoadResult.Data, HasTracks
+    }
+
+    @Serializable
+    data class SearchResult private constructor(
+        override val loadType: ResultStatus,
+        override val data: Data
```
topi314, line 49: can we also just use `List<Track>` here?

DRSchlaubi: That would not implement Data

topi314: oh true lol

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/loadResult.kt
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * Severity levels for FriendlyException
+     */
+    @Serializable
+    enum class Severity {
+        /**
+         * The cause is known and expected, indicates that there is nothing wrong with the library itself.
+         */
+        COMMON,
```
topi314, line 158: these also need the `@SerialName` annotation right?

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/loadResult.kt
```diff
@@ -0,0 +1,177 @@
...
+    @SerialName("track")
+    TRACK,
+
+    @SerialName("playlist")
+    PLAYLIST,
+
+    @SerialName("searchResult")
+    SEARCH_RESULT,
+
+    @SerialName("none")
```
topi314: this should be `empty`

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/loadResult.kt
```diff
@@ -0,0 +1,177 @@
...
+@Serializable
+enum class ResultStatus {
+    @SerialName("track")
+    TRACK,
+
+    @SerialName("playlist")
+    PLAYLIST,
+
+    @SerialName("searchResult")
+    SEARCH_RESULT,
```
topi314: this should be `search`

topi314:
I updated my client to the discussed & here implemented changes and it seems to be working fine

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/messages.kt
```diff
@@ -0,0 +1,252 @@
...
+
+            /**
+             * Reason why a track stopped playing.
+             *
+             * @property mayStartNext Indicates whether a new track should be started on receiving this event.
+             * If this is false, either this event is already triggered because another track started (REPLACED)
+             * or because the player is stopped (STOPPED, CLEANUP).
+             */
+            @Serializable
+            enum class AudioTrackEndReason(val mayStartNext: Boolean) {
```
viztea, line 163: Enum members are serialized with different names. Refer to v4 IMPLEMENTATION.md

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/loadResult.kt
```diff
@@ -0,0 +1,180 @@
...
+@Serializable
+enum class ResultStatus {
+    @SerialName("track")
+    TRACK,
+
+    @SerialName("playlist")
+    PLAYLIST,
+
+    @SerialName("search")
+    SEARCH_RESULT,
```
viztea: Maybe rename this enum member to `SEARCH`?

protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/player.kt
```diff
@@ -0,0 +1,73 @@
...
+    val position: Long,
+    val title: String,
+    val uri: String?,
+    val sourceName: String,
+    val artworkUrl: String?,
+    val isrc: String?
+)
+
+@Serializable
+data class VoiceState(
```
topi314, line 45: they are marked as not optional in the spec

viztea (approved):

Drapersniper:
The Track search response is currently broken.

```json
{
	"loadType": "search",
	"data": {
		"tracks": []
	}
}
```
is what's being returned when it should be 

```json
{
	"loadType": "search",
	"data": []
}
```

DRSchlaubi:
> The Track search response is currently broken.
> 
> ```json
> {
> 	"loadType": "search",
> 	"data": {
> 		"tracks": []
> 	}
> }
> ```
> 
> is what's being returned when it should be
> 
> ```json
> {
> 	"loadType": "search",
> 	"data": []
> }
> ```

fixed

build.gradle.kts
```diff
@@ -54,15 +51,17 @@ subprojects {
...
+    Grgit.open(mapOf("currentDir" to project.rootDir)).use { git ->
+        val headTag = git.tag
+            .list()
+            .find { it.commit.id == git.head().id }
 
-    val clean = git.status().isClean || System.getenv("CI") != null
-    if (!clean) {
-        println("Git state is dirty, setting version as snapshot.")
-    }
+        val clean = git.status().isClean || System.getenv("CI") == null
```
topi314:
```suggestion
        val clean = git.status().isClean || System.getenv("CI") != null
```

topi314 (changes requested):

build.gradle.kts
topi314: not sure why this was changed

PR #853, opened 2023-03-14
Do not register Mixer/Beam source manager

TheEssem:
Mixer was shut down in 2020, making this source manager useless.


ooliver1 (commented):
Does the config example need updating?

TheEssem:
> Does the config example need updating?

No, that was taken care of in 620bd16d91cdd71e70e888966b99a2c5ac706f5d.

topi314:
ty

PR #857, opened 2023-03-28
Rename application.yml.example to application.yml

RUTUJEKWIELKI:


topi314:
No I don't want my `application.yml` for development checked into git

PR #860, opened 2023-04-04
Lavalink.kt supports rest api since 4.0.0

DRSchlaubi:
https://github.com/DRSchlaubi/Lavalink.kt/releases/tag/4.0.0


topi314:
feel free to target the `v4` branch instead

DRSchlaubi:
Lavalink.kt 4.0.0 doesn't target Lavalink v4 it targets v3 and therefore that doesn't make a lot of sense

topi314:
ahh nvm then

PR #865, opened 2023-04-17
use lavadsp fork with darwin arm stuff fixed

topi314:
current lavadsp does not work on the new m1 macs. Since lavadsp is archived again I forked it &  monkey patched it to not use avx2 natives on m1 macs. This resolves the issue.

to note:
I removed freedsb support which isn't supported by lavaplayer anyway and therefore not needed

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -28,6 +28,7 @@ allprojects {
         mavenCentral() // main maven repo
         mavenLocal()   // useful for developing
         maven("https://m2.dv8tion.net/releases")
+        maven("https://maven.topi.wtf/releases")
         jcenter()
         maven("https://jitpack.io") // build projects directly from GitHub
     }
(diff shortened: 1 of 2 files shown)


duncte123:
Please don't remove musl, I use that stuff

topi314:
Lavaplayer doesn't ship with musl natives right
How does that work 🤔

duncte123:
> Lavaplayer doesn't ship with musl natives right How does that work 🤔

I put the natives in my own resources folder at the correct path and it overrides the lavaplayer ones

topi314:
Aaaaaaaa

topi314:
alright this should have working musl support now, feel free to give it a test @duncte123

duncte123:
> alright this should have working musl support now, feel free to give it a test @duncte123

Crash while using filters https://cdn.discordapp.com/attachments/1082305147042013256/1100327677883326494/hs_err_pid1.log

topi314:
@duncte123 I think there was a misunderstanding
We just readded musl support for lavadsp, but not added musl support for lavaplayer. That would mean you need to run the jar with your beautiful dockerfile setup

I have tested it with a jar based on the x86-64 musl libconnector.so you also use

duncte123:
@TopiSenpai that works for me

topi314:
Are you able to give it another test before we consider merging this?

topi314:
closing this in favour of https://github.com/lavalink-devs/Lavalink/pull/878

PR #898, opened 2023-06-29
release v3.7.6

topi314:
## What's Changed
* Update Lavaplayer to [`1.4.1`](https://github.com/Walkyst/lavaplayer-fork/releases/tag/1.4.1) & [`1.4.2`](https://github.com/Walkyst/lavaplayer-fork/releases/tag/1.4.2)
* New support for `MUSL` based systems (most notably `alpine`)
* New `alpine` docker image variant (use `-alpine` suffix)


LavalinkServer/docker/Dockerfile.alpine
freyacodes, line 1: It seems to be convention that this file should be called alpine.Dockerfile. E.g. https://docs.docker.com/compose/compose-file/build/#illustrative-example

topi314: good catch

freyacodes (approved):

PR #914, opened 2023-07-11
Update v4 clients list adding Coglink

ThePedroo:


freyacodes (approved):

topi314:
Not sure why you removed your client from v3.7 support list

PR #940, opened 2023-08-09
Fixed typo in README.md

Fyphen1223:


Fyphen1223:
This will be temp fix @topi314

topi314:
neither does that link resolve to anything & I'm gonna wait for github to fix this
https://github.com/Fyphen1223/Lavalink/tree/lavalink-devs/CHANGELOG.md

Fyphen1223:
I changed link and that should work, and this will be temp fix

topi314:
it does not work, nor do I want a temp fix

Fyphen1223:
k
sry to be annoying

Fyphen1223:
but let me just say this
https://github.com/lavalink-devs/Lavalink/blob/master/CHANGELOG.md#400-beta1
this link is ok right?

topi314:
that link would indeed work

Fyphen1223:
and MD link is this

topi314:
but linking to a static branch is just stupid, github should just fix it

Fyphen1223:
why only this link is not working

like testbot is working properly
plugins.md is ok

topi314:
I recommend asking github support

Fyphen1223:
um that's true enough

PR #970, opened 2023-10-11
Add region balancing note to JLC

duncte123:


topi314 (commented):
I think this is a bit too specific (and long)
This should be a broad overview and nothing more
You are welcome to list all features in the clients README and even compare to other libraries

PR #983, opened 2023-11-26
Allow setting user data on tracks

topi314:
https://github.com/lavalink-devs/Lavalink/pull/984


protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/player.kt
```diff
@@ -23,8 +23,13 @@ data class Player(
 data class Track(
     val encoded: String,
     val info: TrackInfo,
-    val pluginInfo: JsonObject
-) : LoadResult.Data
+    val pluginInfo: JsonObject,
+    val userData: JsonObject
+) : LoadResult.Data {
+    fun copyWithUserData(userData: JsonObject): Track {
```
duncte123, line 67: Maybe it's a good idea to add "deserializeUserData" like was done in #973 for plugin info

protocol/src/commonTest/kotlin/PlayerSerializerTest.kt
```diff
@@ -127,7 +129,7 @@ class PlayerSerializerTest {
     @JsName("test3")
     fun `test encodedTrack and identifier exclusivity`() {
         //language=json
-        val json = """{"encodedTrack":  "", "identifier":  ""}"""
+        val json = """{"track": {"encodedTrack":  "", "identifier":  ""}}"""
```
duncte123: Invalid json property
```suggestion
        val json = """{"track": {"encoded":  "", "identifier":  ""}}"""
```

LavalinkServer/src/main/java/lavalink/server/player/PlayerRestHandler.kt
```diff
@@ -59,8 +59,24 @@ class PlayerRestHandler(
     ): ResponseEntity<Player> {
         val context = socketContext(socketServer, sessionId)
 
-        val encodedTrack = playerUpdate.encodedTrack
-        if (encodedTrack is Omissible.Present && playerUpdate.identifier is Omissible.Present) {
+        val track = if (playerUpdate.track.isPresent()) {
+            playerUpdate.track
+        } else {
```
freyacodes, line 68: We should return 400 Bad Request if both methods of specification are used

freyacodes (changes requested):
Documentation is needed

topi314:
docs are here https://github.com/lavalink-devs/Lavalink/pull/984

freyacodes (approved):

duncte123 (approved):

PR #990, opened 2023-12-03
Add java json API

DRSchlaubi:


duncte123:
It would be a good idea to add serialisation/deserialisation helpers as well

DRSchlaubi:
> It would be a good idea to add serialisation/deserialisation helpers as well

https://github.com/lavalink-devs/Lavalink/blob/67ff059c92464716b24bd2ff6993a054ca519e85/protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/utils/serialization.kt

duncte123 (approved):
Meant objects instead of just strings, but this will do for now

DRSchlaubi:
It allows for serialization of JsonPrimitive, JsonArray and JsonObject

duncte123:
> It allows for serialization of JsonPrimitive, JsonArray and JsonObject

But what about 
```java
public record MyTrackUserData(long userId, String anotherArg) {}
```

Currently it cannot handle this for java users

DRSchlaubi:
> > It allows for serialization of JsonPrimitive, JsonArray and JsonObject
> 
> But what about
> 
> ```java
> public record MyTrackUserData(long userId, String anotherArg) {}
> ```
> 
> Currently it cannot handle this for java users

Well yeah, but kotlinx.serialization has a compiler plugin, which generates the serializer (this is what @Serializable does), you cannot do that for a java record, so I can't handle that

topi314:
replaced by https://github.com/lavalink-devs/lavalink-client/pull/12

PR #991, opened 2023-12-04
fix plugin manager not deleting old plugin version

topi314:


LavalinkServer/src/main/java/lavalink/server/bootstrap/PluginManager.kt
```diff
@@ -35,14 +36,11 @@ class PluginManager(val config: PluginsConfig) {
...
+        data class PluginJar(val manifest: PluginManifest, val file: File)
 
-        val pattern = Pattern.compile("(.+)-(.+)\\.jar$")
-        val pluginJars = directory.listFiles()!!.mapNotNull { f ->
-            val matcher = pattern.matcher(f.name)
-            if (!matcher.find()) return@mapNotNull null
-            PluginJar(matcher.group(1), matcher.group(2), f)
-        }
+        val pluginJars = directory.listFiles()!!.filter { it.extension == "jar" }.map {
+            loadPluginManifests(JarFile(it)).map { manifest -> PluginJar(manifest, it) }
```
duncte123: How does loading the manifests here impact performance compared to the regex?

LavalinkServer/src/main/java/lavalink/server/bootstrap/PluginManager.kt
```diff
@@ -124,18 +126,11 @@ class PluginManager(val config: PluginsConfig) {
...
-        val manifests = mutableListOf<PluginManifest>()
-
-        jar.entries().asIterator().forEach { entry ->
-            if (entry.isDirectory) return@forEach
-            if (!entry.name.startsWith("lavalink-plugins/")) return@forEach
-            if (!entry.name.endsWith(".properties")) return@forEach
-            manifests.add(parsePluginManifest(jar.getInputStream(entry)))
-        }
 
+        val manifests = loadPluginManifests(jar)
```
duncte123: Also mentioned this in discord, but this is probably cachable with the jars that you have already loaded, so you don't have to open the files for a second time.

LavalinkServer/src/main/java/lavalink/server/bootstrap/PluginManager.kt
topi314: I did some quick & dirty testing and it seems to have no impact at all
startup times were always in the same 100ms frame

LavalinkServer/src/main/java/lavalink/server/bootstrap/PluginManager.kt
topi314: I don't think this makes an actual differenc

freyacodes (approved):

PR #998, opened 2023-12-27
Add nyxx_lavalink to clients

abitofevrything:
Adds the `lavalink` and `nyxx_lavalink` clients for Dart to the client list.

`lavalink` is library-agnostic and `nyxx_lavalink` establishes a bridge between it and the `nyxx` library. They are both hosted in a monorepo, which is why the repository link is the same.
Currently the packages are only available on the `feat/nyxx-6.0.0` branch in that repository, but they will be available on `main` once `nyxx_lavalink` v4 comes out of beta.


devoxin:
Given that the repositories are identical, it might be better to just have the one entry and put any additional information you want users to know within the relevant column to avoid bloating the list

topi314:
Ideally you redirect users in your own repository to the other module

abitofevrything:
> Given that the repositories are identical, it might be better to just have the one entry and put any additional information you want users to know within the relevant column to avoid bloating the list.

Done!

PR #1002, opened 2024-01-06
Add LLTurbo client

Fyphen1223:
Add LLTurbo client.

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -19,6 +19,7 @@ description: A list of Lavalink client libraries.
 | [Lavalink-Client](https://github.com/tomato6966/Lavalink-Client)    | Node.js         | **Any**                                    |                                |
 | [FastLink](https://github.com/PerformanC/FastLink)                  | Node.js         | **Any**                                    |                                |
 | [Riffy](https://github.com/riffy-team/riffy)                        | Node.js         | **Any**                                    |                                |
+| [LLTurbo](https://github.com/Fyphen1223/LLTurbo)                    | Node.js         | **Any**                                    |                                |
 | [DisCatSharp](https://github.com/Aiko-IT-Systems/DisCatSharp)       | .NET            | DisCatSharp                                | v10.4.2+                       |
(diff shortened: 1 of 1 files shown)


viztea:
This is literally a fork of Shoukaku with a name change...

Fyphen1223:
Nah it supports voice receiving support on NodeLink...
Should I send this to NodeLink not here?

Fyphen1223:
Um I think so, I will reopen this PR when I changed this client drastically. Sorry for disturbing.

1Lucas1apk:
I think this modification is not valid for adding a package that supports other software that is not lavalink, the right thing to do would be to go to NodeLink and ask them to add it there

ThePedroo:
I totally disagree. NodeLink follows LavaLink protocols. If this is not accepted by that reason, clients that adds a plugin support shouldn't be added since it's not LavaLink.

I didn't look at the source, but if this is only the addition of NodeLink features, it should be a PR to shoukaku, not its own client. Now, if it's an almost entirely new base without being mostly the same, it's okay, as we have erela.js forks that are now clients.

1Lucas1apk:
I disagree, imagine a user making a fork in the coglink repo, and making a modification to a software that has different functionalities and then announcing it as a package with a different name, I feel like it's undervaluing the user's work, and it's not that question, he modified the shoudoku package to with nodelink functionalities, it's the same thing as having 2 shodouku only one with a different name, and functionality that won't work with lavalink, and it's unnecessary to try to put it on the client list, as you said in a certain part, if he at least hired the owner of the package to make it compatible it would also be ideal
Or ask another software to add it as a "nodelink" client

ThePedroo:
Please do not put words in my mouth.

"but if this is only the addition of NodeLink features, it should be a PR to shoukaku"
"not its own client."
"almost entirely new base without being mostly the same, it's okay"

devoxin:
As it currently stands this library isn't distinctive enough to warrant adding to the client library list, therefore it won't be added until it is at a point where it could be considered its own client, rather than just a fork of.

PR #1003, opened 2024-01-26
Add support for Kotlin/Native in protocol

DRSchlaubi:
In preparation of Kord & Lavalink.kt adding support for Kotlin/Native this PR updates the protocol module accordingly

Lavalink.kt & Kord are currently waiting for ktor 3.0 (or at least a beta) to use this, so priority on this isn't high as of rn

**Breaking change** This PR un-ignores the `gradle.properties` file, if you use this file for any private data, please use `~/.gradle/gradle.properties` instead


DRSchlaubi:
Ktor 3.1 released the changed required, would appriciate if we could release this at some point in the near future

topi314:
So, does this still require a mac runner?

DRSchlaubi:
> So, does this still require a mac runner?

Not as of Kotlin 2.1.0: https://kotlinlang.org/docs/whatsnew21.html#how-to-enable-publishing-libraries-from-any-host

DRSchlaubi:
Also macos runners are free for public projects

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.8-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-bin.zip
```
topi314: is this gradle downgrade needed?

gradle.properties
```diff
@@ -0,0 +1,2 @@
+kotlin.native.enableKlibsCrossCompilation=true
```
topi314: can we put this somewhere else?
I'd rather not have the `gradle.properties` committed

topi314 (changes requested):

gradle/wrapper/gradle-wrapper.properties
DRSchlaubi: I assume this happens when merging in master as master has updated the version since

gradle.properties
DRSchlaubi: The gradle.properties is supposed to be committed, ignoring it was wrong in the first place

topi314: Apparently user specific configuration should go into `~/.gradle/gradle.properties`.
Can you add that into the pr description?

topi314:
this needs to be rebased, there are a lot of unrelated changes in here

topi314 (approved):

PR #1042, opened 2024-04-27
Expose lavalink sessions for plugins

topi314:
This pr adds the ability for plugins to easily access the sessions without having to import the lavalink server as a dependency

in some plugins i want to have rest routes such as `/v4/sessions/{SESSION_ID}/players/{GUILD_ID}`

this is rather annoying atm since you have to add the lavalink server as a compile only dependency


topi314:
two such example plugins would be https://github.com/topi314/LavaLyrics or https://github.com/topi314/LavaQueue

devoxin (approved):
LGTM

freyacodes (commented):
This is a good way to implement this. However, it lacks documentation.

PR #1047, opened 2024-05-24
Add fixes to documentation

mplatypus:
This PR fixes the following:

A mistake in a example payload, where it was supposed to be `semver`, but instead it was `string`

Adds a 204 No content on the `/v4/routeplanner/status` endpoint, which is received when the routeplanner is disabled.

I would like to say, that the way it has been written, is not perfect, so let me know if it needs changes.


topi314 (approved):

topi314:
thanks

PR #1054, opened 2024-06-06
Add @discordx/lava-player to clients.md

vijayymmeena:
Website: https://discordx.js.org/docs/lava-player
NPM: https://www.npmjs.com/package/@discordx/lava-player
Repository: https://github.com/discordx-ts/discordx/tree/main/packages/lava-player#readme


devoxin:
After some internal deliberation, we have decided not to accept any new submissions for JS-based, or JS-derived clients due to the increasing size of the client list. We want to enable users to find a client suited for their needs, but we would like to avoid overwhelming them with too much choice. Additionally, as the number of JS-based/derived clients increases, we feel it is our duty to keep the list under a modicum of control to prevent it from growing endlessly.

topi314:
some followup for clarity:
We will reconsider this submission once we have revisited the current list if clients

PR #1058, opened 2024-06-18
Feature/binary compatibility

DRSchlaubi:
Requires #1056


nilscrafthd:
+1

topi314:
i don't think we need that check tbh

PR #1059, opened 2024-06-23
Dependency upgrades

DRSchlaubi:
Replaces #1056


topi314 (approved):
I tested this and it does not seem to break kotlin plugins compiled with kotlin `v1.8.22` or `1.9`

PR #1074, opened 2024-07-13
Update Docs

viztea:
Fills the guild id fields in example payloads with a valid snowflake (lavalink server id) so that it's possible to more easily copy/paste them when writing unit tests.

Also added the user data field to tracks


viztea:
Actually, leave this open for now. When I get around to implementing the REST API, I may need to change those docs too.

PR #1088, opened 2024-08-09
Added Oracle Cloud ipv6 docs configuration

szymonwilczek:


topi314:
this guide also uses tunnelbroker
how does this differ compared to https://lavalink.dev/configuration/ipv6/tunnelbroker.html?
I'd rather not have 2 copies of the same thing
if oracle cloud needs some extra commands then it should rather first redirect to the general tunnelbroker guide

szymonwilczek:
Main configuration in tunnelbroker guide uses example on netplan, which isn't configured as primary network interface for oracle. I thought the creation of images-like tutorial would be cool anyway

topi314:
how about you create a generic non netplan tunnelbroker variant then?
I think that would make more sense to cover as many setups as possible

szymonwilczek:
that would be great. I'll take a look at it when I have some time. i think we can close pr in that case

topi314:
you can keep this pr open and just push changes as you make those, but it's your choice

topi314:
your commits will get squashed into one at the end

szymonwilczek:
when I have some changes ready, i'll reopen again

docs/configuration/ipv6/oraclecloud.md
```diff
@@ -0,0 +1,185 @@
...
+- The pop-up will show:
+  
+![image](https://github.com/user-attachments/assets/641c0883-9437-42fa-9d3a-49f5ac7330a3)
+
+Configure it accordingly:
+- For `Source CIDR` type: **0.0.0.0/0**
+- For `IP Protocol` change to: **ICMP**
+So your fully configured `Ingress Rule` should look like this:
+
+![image](https://github.com/user-attachments/assets/ef467326-05e2-4f3a-ac30-ae33726b66f4)
```
Jackenmen, line 44: Wouldn't allowing the Echo Request ICMP packet (type 8) rather than all be enough? TunnelBroker's FAQ seems to suggest that.

szymonwilczek: Oh jesus, right. I completely forgot to exclude the `All` from the image. There was never meant to be the `All` since it's allowing too much.

PR #1107, opened 2024-10-26
Update springboot to 3.3.5

hwangsihu:


topi314:
Any specific reason you want spring 3.3.5?

We usually bundle dependency updates as one big pr

hwangsihu:
Because there may be bug fixes and stability improvements.
Ah, I tested it and it works fine.

topi314:
There may?
In this case we will bump dependencies as we usually do

hwangsihu:
I don't think it needs to be applied right now, so it would be better to merge it when the next version is released.

hwangsihu:
https://github.com/hwangsihu/Lavalink/commit/9c321349b1e4d82dd52de6236dc0dfae428b4a61
What do you think about this? I haven't found any errors yet.

PR #1112, opened 2024-11-12
fix: remove tls flags from dockerfiles

KagChi:


topi314 (changes requested):
seems like this line was added because TLS 1.3 was broken at some point in java 11.

Just remove it.

https://github.com/lavalink-devs/Lavalink/commit/fd813b0c9d36daba29ad1d9a66fe7e08f9f698a9

PR #1120, opened 2024-12-10
Adds a plugin update checker

### LINKED ISSUE #1044: [Plugins] Automatic Update Checking
### Change request

As suggested in the Lavalink discord, the possibility of having the Lavalink server automatically check for updates using something like `maven-metadata.xml` to identify the latest plugin versions would be very useful, particularly in letting users know there's a new version of a plugin (or plugin***s***) they use which could resolve any potential issues they might be experiencing, as well as introducing features etc. I believe this would be huge, from a user experience perspective.

### Relevance

Lavalink is responsible for loading plugins, sometimes even fetching them if supplied via config, so it makes sense for Lavalink to alert people to new versions if possible. Plugin devs *could* implement their own update checking but this way ensures that as many plugins would benefit from it as possible.

### Could this be a plugin instead?

I don't believe so, unless plugins can be aware of other registered plugins. Although ideally this would be something integrated within the server directly for simplicity's sake and to ensure one global standard and general compliance.

### Checklist

- [X] I have checked for duplicate issues
- [X] I have checked for existing plugins

sh30801:
Adds a plugin update checker function to `PluginManager.kt` that runs on application startup

Closes #1044

The function checks the maven repository of each plugin declaration and searches the maven metadata file for the latest version. If then compares the latest version and current version installed and writes to the console if the user needs to update. The 3rd party library "semver" is used as a dependency to compare versions.


topi314:
your branch appears to have been made from master, can you rebase your changes onto dev?

sh30801:
> your branch appears to have been made from master, can you rebase your changes onto dev?

I already did but it looks like it messed up still. I'll try again

topi314:
worst case, you can `git reset` your commits, commit them again & just force push to your branch

Issue #1049, opened 2024-05-28
[Plugins] Parse `maven-metadata.xml` when downloading


topi314:
### Change request

Right now Lavalink expects the maven artifacts to have the exact name as the `artifact` id.
But this isn't always the case, as an example when publishing a `-SNAPSHOT` release maven replaces `SNAPSHOT` with the current date. This then causes Lavalink to not being able to download the plugin anymore.

The solution is to instead read in the `maven-metadata.xml` which then points to the correct artifact for a certain plugin. What this also means is that you need a propper maven repository with a `maven-metadata.xml` to host Lavalink plugins, which was previously not needed.

Ideally we would implement this as an addition to the current way of doing it, which would allow us to not have to do [this](https://github.com/lavalink-devs/youtube-source/blob/37aceecadbf0e096005d9db0ae1e5a265cbd5ee9/build.gradle.kts#L15) anymore.

### Relevance

yes

### Could this be a plugin instead?

no

### Checklist

- [X] I have checked for duplicate issues
- [X] I have checked for existing plugins

PR #1121, opened 2024-12-10
Add Plugin Update Checker

### LINKED ISSUE #1044: [Plugins] Automatic Update Checking

sh30801:
Adds a plugin update checker function to `PluginManager.kt` that runs on application startup

Closes #1044

The function checks the maven repository of each plugin declaration and searches the maven metadata file for the latest version. If then compares the latest version and current version installed and writes to the console if the user needs to update. The 3rd party library "semver" is used as a dependency to compare versions.

### DIFF
--- a/LavalinkServer/build.gradle.kts
+++ b/LavalinkServer/build.gradle.kts
@@ -8,6 +8,7 @@ plugins {
     application
     kotlin("jvm")
     id("org.jetbrains.dokka")
+    kotlin("plugin.serialization")
     alias(libs.plugins.maven.publish.base)
 }
 
@@ -30,6 +31,15 @@ java {
     sourceCompatibility = JavaVersion.VERSION_17
 }
 
+kotlin {
+    sourceSets {
+        all {
+            languageSettings.optIn("kotlinx.serialization.ExperimentalSerializationApi")
+            languageSettings.optIn("nl.adaptivity.xmlutil.ExperimentalXmlUtilApi")
+        }
+    }
+}
+
 configurations {
     compileOnly {
         extendsFrom(annotationProcessor.get())
@@ -41,6 +51,9 @@ dependencies {
     implementation(projects.pluginApi) {
         exclude(group = "org.springframework.boot", module = "spring-boot-starter-tomcat")
     }
+
+    implementation(libs.xmlutil.jdk)
+    implementation(libs.xmlutil.serialization)
 
     implementation(libs.bundles.metrics)
     implementation(libs.bundles.spring) {
--- a/LavalinkServer/build.gradle.kts
+++ b/LavalinkServer/build.gradle.kts
@@ -65,6 +65,7 @@ dependencies {
     implementation(libs.kotlin.reflect)
     implementation(libs.logback)
     implementation(libs.sentry.logback)
+    implementation(libs.semver)
     implementation(libs.oshi) {
         // This version of SLF4J does not recognise Logback 1.2.3
         exclude(group = "org.slf4j", module = "slf4j-api")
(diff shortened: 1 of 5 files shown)


LavalinkServer/build.gradle.kts
```diff
@@ -65,6 +65,7 @@ dependencies {
     implementation(libs.kotlin.reflect)
     implementation(libs.logback)
     implementation(libs.sentry.logback)
+    implementation(libs.semver)
```
topi314: is this lib really needed? we already have a way to parse semver in https://github.com/lavalink-devs/Lavalink/blob/88cc85978198f75b662990b1ed0b0e9c77cd347b/protocol/src/commonMain/kotlin/dev/arbjerg/lavalink/protocol/v4/info.kt#L59

you would just need to add a method to compare them which should be easy

LavalinkServer/src/main/java/lavalink/server/bootstrap/PluginManager.kt
```diff
@@ -83,6 +85,37 @@ class PluginManager(val config: PluginsConfig) {
...
+
+                    if(latestVersion > currentVersion) {
+                        log.warn("A newer version of ${declaration.name} was found: $latestVersion. " +
+                                "The current version is $currentVersion.")
+                    } else {
+                        log.info("Plugin ${declaration.name} is up to date")
+                    }
+                    break
+                }
+            }
```
topi314: Can you please not regex XML?
With #1049 we want to parse the `maven-metadata.xml` anyway

topi314 (changes requested):

LavalinkServer/build.gradle.kts
sh30801: Oh I didnt see this thanks!

LavalinkServer/src/main/java/lavalink/server/bootstrap/PluginManager.kt
```diff
@@ -9,9 +9,12 @@ import java.io.FileOutputStream
 import java.io.InputStream
 import java.net.URL
 import java.net.URLClassLoader
+import java.net.HttpURLConnection
 import java.nio.channels.Channels
 import java.util.*
 import java.util.jar.JarFile
+import javax.xml.parsers.DocumentBuilderFactory
```
topi314: wasn't this removed on java 11?
https://www.oracle.com/java/technologies/javase/11-relnote-issues.html#JDK-8190378

either [jackson xml](https://github.com/FasterXML/jackson-dataformat-xml) or [kotlinx.serialization](https://github.com/pdvrieze/xmlutil) should do the trick here

sh30801: no that looks like it only affects the java namespace, javax looks like its untouched by that

topi314: I'd rather have this use a modern lib like the above

topi314 (approved):
I went ahead and changed it to pdvrieze xmlutil, thanks

PR #1135, opened 2025-03-05
Adding a `User-Agent` header to socket connections 

holasoyender:
This is a suggestion that was discussed on the Discord support server. Related to https://github.com/lavalink-devs/lavalink-client/pull/34

This pull request adds the option for clients to set a `User-Agent` header on socket connections to set extra information about the bot.

### Headers example
```
Authorization: youshallnotpass
User-Id: 170939974227541168
Client-Name: lavalink-client/2.0.0
User-Agent: TestBot/1.0.0
```

When the `User-Agent` header is set, every log will have a prefix to identify the client its coming from.


holasoyender:
As there is no actual way of creating a relation between a websocket connection and a http request, the easiest way of implementing it is to read the user-agent from the requests, asking for client library devs to also include that header on every request.

Related to https://github.com/holasoyender/lavalink-client/commit/8dfaf59bcb539ff99c15754a673841c320c60ea1

PR #1141, opened 2025-03-23
fixed filter values being omitted when default value is used

topi314:


docs/api/rest.md
```diff
@@ -401,95 +401,95 @@ where -0.25 means the given band is completely muted, and 0.25 means it is doubl
...
 </details>
 
-| Field | Type  | Description             |
-|-------|-------|-------------------------|
-| band  | int   | The band (0 to 14)      |
-| gain  | float | The gain (-0.25 to 1.0) |
+| Field | Type  | Default | Description             |
+|-------|-------|---------|-------------------------|
+| band  | int   |         | The band (0 to 14)      |
+| gain  | float | 1.0     | The gain (-0.25 to 1.0) |
```
devoxin: Is it worth changing the default? 1.0 is quite a large change and in most cases would cause distortion. I perhaps wouldn't have a default in this case. Also might be worth documenting that 0.0 means gain is unchanged, as in most other cases `1.0` means unchanged which isn't true for equalizer.

topi314: I don't think it's worth since it's always been `1.0`
that would technically be a breaking change

devoxin: It seems more of a quality-of-life change but I guess if anything it's probably worth not documenting defaults for filters like these, where both properties are basically required because defaults aren't really good for anything

topi314: technically they aren't even marked as optional

perhaps we should just get rid of the default values
but then, that would be breaking as someone might rely on this undocumented behavior and I'd rather not break that

then ig we don't really need to document them since it's "undocumented"

PR #1144, opened 2025-04-16
 add support for spring cloud config & docs 

topi314:


infnibor:
## Suggestions:
- Add documentation under the endpoint for example /docs 
- Add a checker, which in the form of e.g. web interface would check if the latest plugin versions are being used and/or Lavalink server data statistics (using Lavalink's REST API)
- Add the possibility to use only the native option - without the need to use the Git repository
- Change the password name to a token or add a comment to direct the user what to enter there correctly

## To check or fix:

~~- Check the reported problem in the conversation regarding a possible problem with plugins outside of Maven (https://discord.com/channels/1082302532421943407/1361868811451502682/1365777795740860527)~~

topi314:
> Add documentation under the endpoint for example /docs

linking to our docs under `/` should be enough

> Add a checker, which in the form of e.g. web interface would check if the latest plugin versions are being used and/or Lavalink server data statistics (using Lavalink's REST API)

the lavalink config server doesnt know about your lavalink servers and it should'nt know anyway. If anything we could try parsing the configs in the config server & check the plugin versions from there

> Add the possibility to use only the native option - without the need to use the Git repository

this should already work, just needs docs

> Change the password name to a token or add a comment to direct the user what to enter there correctly

done

PR #1151, opened 2025-05-27
Rework CPU stats reporting 

### LINKED ISSUE #1132: Invalid lavalink load when many connected client
## Description
The lavalink load and server load become incorrect (0) when many clients connect at the same time.
The load should be calculated 1 time every 30s, not every time it's requested.

## Expected behavior
A correct value

Output of `java -version`:
```
➜  ~ java --version
openjdk 17.0.13 2024-10-15
OpenJDK Runtime Environment (build 17.0.13+11-Debian-2deb12u1)
OpenJDK 64-Bit Server VM (build 17.0.13+11-Debian-2deb12u1, mixed mode, sharing)
➜  ~
```

kikkia:
Fix #1132 

I reworked how we report cpu loads. Now they will only be calculated once ever 30 seconds (easy to change if wanted). 
Previously, the process CPU calculation could be unreliable if called too frequently.

The core logic of the calculations is largely unchanged bar some renaming to make stuff more descriptive.

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/StatsCollector.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/StatsCollector.kt
@@ -28,6 +28,7 @@ import org.slf4j.LoggerFactory
 import org.springframework.web.bind.annotation.GetMapping
 import org.springframework.web.bind.annotation.RestController
 import oshi.SystemInfo
+import java.util.concurrent.locks.ReentrantLock
 import kotlin.Exception
 
 @RestController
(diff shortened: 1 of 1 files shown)


LavalinkServer/src/main/java/lavalink/server/io/StatsCollector.kt
```diff
@@ -40,34 +41,94 @@ class StatsCollector(val socketServer: SocketServer) {
...
+    private fun getOrUpdateCpuStats(): Cpu {
+        val currentCachedCpu = cachedCpu
+        val currentLastCalcTime = lastCpuCalcTime
+
+        if (currentCachedCpu != null && (System.currentTimeMillis() - currentLastCalcTime <= CPU_STATS_REFRESH_INTERVAL_MS)) {
+            return currentCachedCpu
+        }
+
+        // Cache miss or stale, so update
+        cpuStatsCalculationLock.lock()
```
devoxin: Why not use synchronised with double check?

kikkia: You mean something like?

```kotlin
val currentCached = cachedCpu
    val currentLastCalc = lastCpuCalcTime
    if (currentCached != null && (System.currentTimeMillis() - currentLastCalc <= CPU_STATS_REFRESH_INTERVAL_MS)) {
        return currentCached
    }

    synchronized(lock) {
        val lockedCached = cachedCpu
        val lockedLastCalc = lastCpuCalcTime
        if (lockedCached == null || (System.currentTimeMillis() - lockedLastCalc > CPU_STATS_REFRESH_INTERVAL_MS)) {
            val newCpu = performCpuStatsCalculation()
            cachedCpu = newCpu
            lastCpuCalcTime = System.currentTimeMillis() 
            return newCpu
        }
        return lockedCached
    }
```

I chose to use an explicit lock due to it being generally more straightforward and readable (imo) whereas using just synchronized blocks is a bit more vague as whats going on and isn't worth the very slight performance bump at the tradeoff of ease of readability/maintainability from my experience.

devoxin: Pretty much. My suggestion is less about performance—to me, synchronised blocks are more readable and the JVM basically handles all of it for you

kikkia: I do like code that is much more explicit in what it's doing, but thats just a personal preference. It's not really a complicated task this, so it's not really a hill worth dying on lol. Ill cook something up with that in a sec

devoxin: Fair, figured I would mention it just as an alternative

LavalinkServer/src/main/java/lavalink/server/io/StatsCollector.kt
```diff
@@ -40,34 +40,90 @@ class StatsCollector(val socketServer: SocketServer) {
...
+    private val cpuStatsCalculationLock = Any()
 
-    // Record for next invocation
-    private val processRecentCpuUsage: Double
-        get() {
-            val p = os.getProcess(os.processId)
-            if (p == null) {
-                log.warn("Failed to get process stats. Process was null.")
-                return 0.0
+    private val CPU_STATS_REFRESH_INTERVAL_MS = 30000
```
topi314: consts should go into the `companion object`

LavalinkServer/src/main/java/lavalink/server/io/StatsCollector.kt
```diff
@@ -40,34 +40,90 @@ class StatsCollector(val socketServer: SocketServer) {
...
+
+    // Baseline for cpu stats calcs
+    private var prevUpTimeMs: Long = 0L
+    private var prevCpuTimeMs: Long = 0L
+
+    /**
+     * Gets or updates the cached CPU stats, depending on last update and cpu stats refresh interval
+     */
+    private fun getOrUpdateCpuStats(): Cpu {
```
topi314, line 62: isn't this whole method still technically a race condition?
since you access `cachedCpu` & `lastCpuCalcTime` outside the lock?

I would prob either put all code behind the lock or just mark the method with `synchronized`. Then you don't have to deal with lock shenanigans

kikkia: This is a pattern known as double checked locking https://en.wikipedia.org/wiki/Double-checked_locking Where we check the locking conditions prior to acquiring the lock as well as after acquiring the lock. Along with the use of Volatile variables it will result in no issues with accesses between threads. However we do need to be careful that for variables being written to within the synchronized code we need the `@Volatile`  annotation present.

topi314 (changes requested):
this should target dev
rebase your changes please

flav-code:
Thank you @kikkia

LavalinkServer/src/main/java/lavalink/server/io/StatsCollector.kt
```diff
@@ -40,34 +40,91 @@ class StatsCollector(val socketServer: SocketServer) {
         private val os get() = si.operatingSystem
 
         private var prevTicks: LongArray? = null
+        private val ticksLock = Any()
+
+        private val CPU_STATS_REFRESH_INTERVAL_MS = 30000
```
topi314:
```suggestion
        private const val CPU_STATS_REFRESH_INTERVAL_MS = 30000
```

also not sure why `si`,`hal`,`os`, `prevTicks` & `ticksLock` were in the compation object to begin with
I think we can move those out? git blame prob goes back to me lmao

topi314 (approved):

PR #1154, opened 2025-06-08
Update information for Lavacord client

MrJacz:
This pull request updates the documentation for Lavalink client libraries in `docs/clients.md`. The changes clarify the compatibility and feature support for the `Lavacord` library across different environments.

Updates to Lavalink client libraries documentation:

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -4,34 +4,34 @@ description: A list of Lavalink client libraries.
 
 # Client Libraries
 
-| Client                                                              | Platform        | Compatible With                            | Additional Information             |
-|---------------------------------------------------------------------|-----------------|--------------------------------------------|------------------------------------|
-| [Lavalink-Client](https://github.com/lavalink-devs/Lavalink-Client) | Java/Kotlin/JVM | JDA/Discord4J/**Any**                      | Uses reactor                       |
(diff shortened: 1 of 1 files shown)


topi314:
If your library supports any, **Any** is enough if you ask me, no need for a long list of libraries

Not sure why the mention of CJS & ESM is needed.

Wouldn't typescript support be better mentioned under platform?

MrJacz:
Thanks for your feedback! You’re right—Lavacord is a lightweight client and doesn’t depend on any specific Discord library. However, it does have built-in wrappers for the listed libraries, so it works out of the box with those. I noticed that other clients list specific library support (plus “Any”), so I followed that style to help users quickly see which libraries work out of the box.

Regarding module support, I included CJS and ESM to highlight that Lavacord provides builds for both formats, since not every client supports both. I agree my original wording could have been clearer—“TS, CJS, and ESM support” is probably the best way to phrase it, as it accurately reflects support for TypeScript source, CommonJS, and ESM builds.

I was thinking of using something like `Written in TS, builds for CJS & ESM` to keep things concise and clear for end users. Just trying to make everything as straightforward as possible—let me know what you think! Ultimately, it’s your call.

topi314:
idk why I need to read chatgpt a response to this.

if you want to mention typescript support the better place would be the platform just like we do it with java/kotlin/jvm if you ask me

MrJacz:
Sorry for copilot response was on mobile and the way i explained what i meant was terrible so used copilot to fix it, I'll change it to that as that makes sense

topi314 (approved):

PR #1178, opened 2025-10-11
add check for empty plugin files and redownload/delete them

kikkia:
https://discord.com/channels/1082302532421943407/1426606307468640256

Seemed easy to just add a basic check for this. Not sure if it's common enough to warrant its own code though, feel free to close if it feels needless

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/bootstrap/PluginManager.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/bootstrap/PluginManager.kt
@@ -133,6 +133,23 @@ class PluginManager(val config: PluginsConfig) {
         Channels.newChannel(URL(url).openStream()).use {
             FileOutputStream(output).channel.transferFrom(it, 0, Long.MAX_VALUE)
         }
+
+        if (output.length() == 0L) {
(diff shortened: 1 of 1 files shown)


LavalinkServer/src/main/java/lavalink/server/bootstrap/PluginManager.kt
```diff
@@ -133,6 +133,23 @@ class PluginManager(val config: PluginsConfig) {
         Channels.newChannel(URL(url).openStream()).use {
             FileOutputStream(output).channel.transferFrom(it, 0, Long.MAX_VALUE)
         }
+
+        if (output.length() == 0L) {
+            log.warn("Downloaded plugin is empty, re-downloading...")
+            if (!output.delete()) {
+                throw RuntimeException("Failed to delete empty plugin file: ${output.path}")
+            }
+            Channels.newChannel(URL(url).openStream()).use {
```
devoxin: Might be better to just add an `isRetry` parameter and reinvoke the function with that set to `true` rather than duplicating logic

devoxin (approved):

topi314 (approved):

PR #1185, opened 2025-12-29
Fix concurrency issue in SocketServer and address deprecations

ramkrishna-xyz:
This PR replaces `mutableMapOf` with `ConcurrentHashMap` for `resumableSessions` in `SocketServer.kt`. This prevents `ConcurrentModificationException` during session resumption and cleanup. It also adds `@Suppress("DEPRECATION")` to `PlayerRestHandler` to reduce build noise.

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/SocketServer.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/SocketServer.kt
@@ -51,7 +51,7 @@ final class SocketServer(
 
     // sessionID <-> Session
     override val sessions = ConcurrentHashMap<String, SocketContext>()
-    override val resumableSessions = mutableMapOf<String, SocketContext>()
+    override val resumableSessions = ConcurrentHashMap<String, SocketContext>()
     private val koe = Koe.koe(koeOptions)
     private val statsCollector = StatsCollector(this)
     private val charPool = ('a'..'z') + ('0'..'9')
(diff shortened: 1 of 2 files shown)


ghost:
I see no point in this at all.

pryzmian:
I feel like this was made with AI

ramkrishna-xyz:
@pryzmian  what made you think so

topi314:
do you have an actual instance of a `ConcurrentModificationException` happening?

ghost:
he probably doesn't lmao

EvilG-MC:
It's probably made entirely with AI ☠️

ramkrishna-xyz:
While we haven't seen an actual `ConcurrentModificationException` crash in the logs yet, I went ahead and fixed the `resumableSessions` map anyway. 
The original code was using a standard `mutableMapOf`, which isn't thread-safe. Since both the WebSocket thread (handling connections) and the background executor (handling session timeouts) touch that same map, it was a guaranteed race condition waiting to happen. I've swapped it to a `ConcurrentHashMap` to make it thread-safe and prevent those random crashes before they actually hit us in production.

ghost:
My face when I use AI:

ToddyTheNoobDud:
> While we haven't seen an actual `ConcurrentModificationException` crash in the logs yet, I went ahead and fixed the `resumableSessions` map anyway.
> The original code was using a standard `mutableMapOf`, which isn't thread-safe. Since both the WebSocket thread (handling connections) and the background executor (handling session timeouts) touch that same map, it was a guaranteed race condition waiting to happen. I've swapped it to a `ConcurrentHashMap` to make it thread-safe and prevent those random crashes before they actually hit us in production.

"chatgpt what is my name" ahh

image

ramkrishna-xyz:
@ToddyTheNoobDud the message was indeed ai generated because my english is like very bad so we will be in a serious type discussion here so I used ai to enhance my pull request message

PR #1188, opened 2026-01-22
feat: add prometheus metrics for load results and track plays

Moquent:
Add new prometheus metrics to track search and play activity by source:
- `lavalink_tracks_played_total{source, guildId}` - counts tracks played by source and guildId
- `lavalink_track_load_results_total{source,result}` counts load result types like search, track, playlist, no_matches, error matched with source

The `source` label is fetched from the load result when available and falls back to parsing from identifier.

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
@@ -9,12 +9,6 @@ import org.springframework.stereotype.Component
 class SearchMetrics {
     
     companion object {
-        private val searchCounter: Counter = Counter.build()
-            .name("lavalink_track_loads_total")
-            .help("Total number of tracks searched by source")
-            .labelNames("source")
-            .register()
-
         private val playCounter: Counter = Counter.build()
             .name("lavalink_tracks_played_total")
             .help("Total number of tracks played by source")
@@ -28,10 +22,6 @@ class SearchMetrics {
             .register()
     }
     
-    fun recordSearch(source: String) {
-        searchCounter.labels(source).inc()
-    }
-
     fun recordPlay(source: String, guildId: String) {
         playCounter.labels(source, guildId).inc()
     }
--- a/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
@@ -1,0 +1,33 @@
+package lavalink.server.metrics
+
+import io.prometheus.client.Counter
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+
+@Component
+@ConditionalOnProperty("metrics.prometheus.enabled")
+class SearchMetrics {
+    
+    companion object {
+        private val searchCounter: Counter = Counter.build()
+            .name("lavalink_searches_total")
+            .help("Total number of tracks searched by source")
+            .labelNames("source")
+            .register()
+
+        private val playCounter: Counter = Counter.build()
+            .name("lavalink_tracks_played_total")
+            .help("Total number of tracks played by source")
+            .labelNames("source")
+            .register()
+    }
+    
+    fun recordSearch(source: String) {
+        searchCounter.labels(source).inc()
+    }
+
+    fun recordPlay(source: String) {
+        playCounter.labels(source).inc()
+    }
+}
+
--- a/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
@@ -10,7 +10,7 @@ class SearchMetrics {
     
     companion object {
         private val searchCounter: Counter = Counter.build()
-            .name("lavalink_searches_total")
+            .name("lavalink_track_loads_total")
             .help("Total number of tracks searched by source")
             .labelNames("source")
             .register()
@@ -18,7 +18,7 @@ class SearchMetrics {
         private val playCounter: Counter = Counter.build()
             .name("lavalink_tracks_played_total")
             .help("Total number of tracks played by source")
-            .labelNames("source")
+            .labelNames("source", "guild_id")
             .register()
     }
     
@@ -26,8 +26,8 @@ class SearchMetrics {
         searchCounter.labels(source).inc()
     }
 
-    fun recordPlay(source: String) {
-        playCounter.labels(source).inc()
+    fun recordPlay(source: String, guildId: String) {
+        playCounter.labels(source, guildId).inc()
     }
 }
 
--- a/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
@@ -20,6 +20,12 @@ class SearchMetrics {
             .help("Total number of tracks played by source")
             .labelNames("source", "guild_id")
             .register()
+
+        private val loadResultCounter: Counter = Counter.build()
+            .name("lavalink_track_load_results_total")
+            .help("Total number of track load results by source and result type")
+            .labelNames("source", "result")
+            .register()
     }
     
     fun recordSearch(source: String) {
@@ -28,6 +34,10 @@ class SearchMetrics {
 
     fun recordPlay(source: String, guildId: String) {
         playCounter.labels(source, guildId).inc()
+    }
+
+    fun recordLoadResult(source: String, result: String) {
+        loadResultCounter.labels(source, result).inc()
     }
 }
 
--- a/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
@@ -16,8 +16,8 @@ class SearchMetrics {
             .register()
 
         private val loadResultCounter: Counter = Counter.build()
-            .name("lavalink_track_load_results_total")
-            .help("Total number of track load results by source and result type")
+            .name("lavalink_item_load_results_total")
+            .help("Total number of audio item load results by source and result type")
             .labelNames("source", "result")
             .register()
     }
--- a/LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
@@ -86,7 +86,7 @@ class AudioLoaderRestHandler(
 
             else -> {
                 log.error("Unknown item type: ${item.javaClass}")
-                searchMetrics?.recordLoadResult("unknown", "unknown")
+                searchMetrics?.recordLoadResult("", "unknown")
                 throw ResponseStatusException(
                     HttpStatus.INTERNAL_SERVER_ERROR,
                     "Identifier returned unknown audio item type: ${item.javaClass.canonicalName}"
@@ -130,7 +130,7 @@ class AudioLoaderRestHandler(
         }
 
         return if (source.isNullOrBlank()) {
-            "unknown"
+            ""
         } else {
             source
         }
--- a/LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
@@ -125,32 +125,34 @@ class AudioLoaderRestHandler(
     }
 
     private fun extractSourceType(item: Any?, identifier: String): String {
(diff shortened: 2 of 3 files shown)


LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
```diff
@@ -116,4 +119,24 @@ class AudioLoaderRestHandler(
             decodeTrack(audioPlayerManager, it).toTrack(it, pluginInfoModifiers)
         }))
     }
+
+    private fun extractSourceType(identifier: String): String {
```
topi314: handling all links by identifier is kinda misleading imo.
if you load a yt link it would get logged as http source which isn't really true
maybe instead we can look at the load result?
this does come with its own issues tho since playlists do not have a source attribute as an example

LavalinkServer/src/main/java/lavalink/server/player/PlayerRestHandler.kt
```diff
@@ -23,6 +23,7 @@ class PlayerRestHandler(
     private val filterExtensions: List<AudioFilterExtension>,
     private val pluginInfoModifiers: List<AudioPluginInfoModifier>,
     serverConfig: ServerConfig,
+    private val searchMetrics: lavalink.server.metrics.SearchMetrics? = null
```
topi314: you can import things

LavalinkServer/src/main/java/lavalink/server/player/PlayerRestHandler.kt
```diff
@@ -213,6 +214,7 @@ class PlayerRestHandler(
                 }
 
                 player.play(newTrack)
+                searchMetrics?.recordPlay(newTrack.sourceManager?.sourceName ?: "unknown")
```
topi314: maybe we could include some additional metrics here like guild id?
that should make the data way more useful imo

topi314 (changes requested):

LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
```diff
@@ -0,0 +1,33 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+
+@Component
+@ConditionalOnProperty("metrics.prometheus.enabled")
+class SearchMetrics {
+    
+    companion object {
+        private val searchCounter: Counter = Counter.build()
+            .name("lavalink_searches_total")
```
topi314: perhaps rename this to something like `lavalink_track_loads_total`.
this would better align the with the naming

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
Moquent: okay yeah, I'm gonna extract the source after load is processed, maybe fallback to identifier if load didn't work. the tracks in the playlist when loaded would have the source attributed though right? something like playlists.tracks.first()?.sourceManager?.sourceName

topi314: there is no guarantee that a playlist only contains tracks from one source, but usually it's the case

Moquent: you're right. for now I used the first track's source name. do you have any other ideas you would want to use instead of this?

topi314:
Considering now you track the load result, it would be useful to track the other result types too

Moquent:
> Considering now you track the load result, it would be useful to track the other result types too

Sure. Do you mean load result outcomes? Like track, playlist, load fail, etc.

topi314:
yes

LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+
+@Component
+@ConditionalOnProperty("metrics.prometheus.enabled")
+class SearchMetrics {
+    
+    companion object {
+        private val searchCounter: Counter = Counter.build()
+            .name("lavalink_track_loads_total")
```
topi314: I don't see how its any useful to have both `lavalink_track_loads_total` and `lavalink_track_load_results_total`

This is duplicate data imo

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
```diff
@@ -116,4 +123,36 @@ class AudioLoaderRestHandler(
             decodeTrack(audioPlayerManager, it).toTrack(it, pluginInfoModifiers)
         }))
     }
+
+    private fun extractSourceType(item: Any?, identifier: String): String {
```
topi314: passing in the identifier here is pointless too imo

topi314 (changes requested):

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
Moquent: sourceName exists only when you actually get an AudioTrack (or a playlist with at least one track). For no_matches, load_failed, or empty playlists, there’s no source returned right? Also some HTTP links that resolve to no item won’t have a track source. So an identifier fallback to atleast show the metric that we got, for example, a no_matches type where we would return unknown source, would be better to return the bare minimum information we have of the type of source / search we used that caused it.

Please correct me if I'm missing something here.

topi314: The information is not bare minimum 
It's just misleading 
Most things you load are http links
So basically everything would be shown as source http which just isnt true

Searches would also show up wrong since `ytsearch` would be converted to `yt`

It's not consistent at all

Moquent: Sorry about that, I see your point. I removed the identifier fallback so source only comes from the load result.

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
```diff
@@ -61,7 +61,7 @@ class AudioLoaderRestHandler(
             loadAudioItem(audioPlayerManager, identifier)
         } catch (ex: FriendlyException) {
             log.error("Failed to load track for identifier $identifier", ex)
-            searchMetrics?.recordLoadResult(extractSourceType(null, identifier), "load_failed")
+            searchMetrics?.recordLoadResult(extractSourceType(null), "load_failed")
```
topi314: no reason to call `extractSourceType` when you pass only null to it
load result type should be `error` here

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
Moquent: Also, since load failed and error results always return "unknown" for the source label, do I just hardcode it into the metric recording?

topi314: if you ask me, the label should just be omitted in that case

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
```diff
@@ -86,15 +86,15 @@ class AudioLoaderRestHandler(
 
             else -> {
                 log.error("Unknown item type: ${item.javaClass}")
-                searchMetrics?.recordLoadResult(extractSourceType(item, identifier), "error")
+                searchMetrics?.recordLoadResult(extractSourceType(item), "error")
```
topi314: this is a weird case here, maybe we should use result type `unknown` here or something
in theory this should never happen anyway but source can return anything they want so it may lol

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
```diff
@@ -86,15 +86,15 @@ class AudioLoaderRestHandler(
...
                 throw ResponseStatusException(
                     HttpStatus.INTERNAL_SERVER_ERROR,
                     "Identifier returned unknown audio item type: ${item.javaClass.canonicalName}"
                 )
             }
         }
 
-        val source = extractSourceType(item, identifier)
+        val source = extractSourceType(item)
```
topi314: just inline this

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
```diff
@@ -59,37 +61,41 @@ class AudioLoaderRestHandler(
...
         } catch (ex: FriendlyException) {
             log.error("Failed to load track for identifier $identifier", ex)
+            searchMetrics?.recordLoadResult(extractSourceType(null), "load_failed")
             return ResponseEntity.ok(LoadResult.loadFailed(ex))
         }
 
-        val result = when (item) {
-            null -> LoadResult.NoMatches()
+        val (result, resultType) = when (item) {
+            null -> LoadResult.NoMatches() to "no_matches"
```
topi314: the result type is named `empty` in the lavalink api

topi314 (changes requested):

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
Moquent: Do you mean just the label (if we do this then it has to be replaced with something because I'm not sure if prometheus will allow not passing a value on a defined label) or the recording of the event itself?

topi314: seems like thats the case
can we pass an empty string? lol

Moquent: yeah that works I guess, did that

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
```diff
@@ -116,4 +121,18 @@ class AudioLoaderRestHandler(
...
+
+    private fun extractSourceType(item: Any?): String {
+        val source = when (item) {
+            is AudioTrack -> item.sourceManager?.sourceName
+            is AudioPlaylist -> item.tracks.firstOrNull()?.sourceManager?.sourceName
+            else -> null
+        }
+
+        return if (source.isNullOrBlank()) {
+            "unknown"
```
topi314: there is still unknown

LavalinkServer/src/main/java/lavalink/server/player/PlayerRestHandler.kt
```diff
@@ -213,6 +215,10 @@ class PlayerRestHandler(
                 }
 
                 player.play(newTrack)
+                searchMetrics?.recordPlay(
+                    newTrack.sourceManager?.sourceName ?: "unknown",
```
topi314: here too

LavalinkServer/src/main/java/lavalink/server/player/AudioLoaderRestHandler.kt
Moquent: oh okay, I didn't know if you wanted to remove the unknown label for all of them. done.

Moquent:
Just checked the new DAVE support stuff, afaik this doesn't break with it. Anything you would want me to do wrt it?

LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
```diff
@@ -0,0 +1,33 @@
...
+    companion object {
+        private val playCounter: Counter = Counter.build()
+            .name("lavalink_tracks_played_total")
+            .help("Total number of tracks played by source")
+            .labelNames("source", "guild_id")
+            .register()
+
+        private val loadResultCounter: Counter = Counter.build()
+            .name("lavalink_track_load_results_total")
+            .help("Total number of track load results by source and result type")
```
topi314: should probably clarify not all load results are tracks

topi314 (changes requested):

LavalinkServer/src/main/java/lavalink/server/metrics/SearchMetrics.kt
Moquent: Good catch, updated to clarify this metric tracks audio item load results, not only tracks.

topi314 (changes requested):
please rebase your changes to dev

topi314 (approved):

PR #1190, opened 2026-02-22
feat: DAVE Support

topi314:


freyacodes:
This change contains a breaking change. Can we release it as version 5.0.0?

PR #1199, opened 2026-03-20
Update Dart's lavalink client support for DAVE

abitofevrything:


docs/clients.md
```diff
@@ -27,6 +27,7 @@ description: A list of Lavalink client libraries.
 | [Coglink](https://github.com/PerformanC/Coglink)                    | C                  | Concord                                    | ✅            |                                    |
 | [Anchorage](https://github.com/Deivu/Anchorage)                     | Rust               | **Any**                                    | ✅            | `tokio`-based                      |
 | [lavalink-rs](https://gitlab.com/vicky5124/lavalink-rs)             | Rust, Python       | **Any**                                    | ✅            | `tokio`-based, `asyncio`-based     |
+| [lavalink](https://github.com/nyxx-discord/nyxx_lavalink)           | Dart               | nyxx/**Any**                               | ✅            |                                    |
```
topi314: can we rename this to `nyxx_lavalink`?

abitofevrything: `nyxx_lavalink` is our client-specific stub that builds upon our client agnostic library that does the actual interacting with the Lavalink API - which is just called `lavalink`. I'm not against listing either one here, but the actual library for the Lavalink API specifically is called `lavalink`.

topi314: if I am not mistaken you are linking to your stub here?
just makes it kinda weird to have a client listed named `lavalink`

edit:
ah they live in the same repo, I'd just name it the same as the github repo makes it more clear

abitofevrything: My only concern is listing the support for client-agnostic usage - in general Dart users don't actually check GitHub that much and just check pub.dev (the Dart package manager), and if you check `nyxx_lavalink` on there directly it isn't at all client agnostic.

Any way we could somehow list both? Or should we just list `nyxx_lavalink` and let people figure out on their own that there also exists a client-agnostic `lavalink` library?

topi314: `nyxx/**Any**` the any already should convey that no?
perhaps you can better highlight in your readme

topi314 (approved):

PR #1200, opened 2026-03-21
Use ConcurrentHashMap for resumableSessions

Sim-hu:
`resumableSessions` in `SocketServer` is currently backed by a plain `mutableMapOf()` (i.e. `LinkedHashMap`), but it's accessed from multiple threads:

- WebSocket handler threads call `remove`/`put` in `afterConnectionEstablished` and `afterConnectionClosed`
- The scheduled executor from `SocketContext.pause()` calls `onSessionResumeTimeout` which does `remove`
- `canResume` reads it from the handshake interceptor thread

This is a race condition that can silently corrupt the map's internal state or throw a `ConcurrentModificationException` at runtime. The `sessions` map right above it already uses `ConcurrentHashMap` for the same reason, so this just makes `resumableSessions` consistent.

One-line change, no new dependencies since `ConcurrentHashMap` is already imported.

### DIFF
--- a/LavalinkServer/src/main/java/lavalink/server/io/SocketServer.kt
+++ b/LavalinkServer/src/main/java/lavalink/server/io/SocketServer.kt
@@ -51,7 +51,7 @@ final class SocketServer(
 
     // sessionID <-> Session
     override val sessions = ConcurrentHashMap<String, SocketContext>()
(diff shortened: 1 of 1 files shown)


topi314:
Duplicate of: #1185

PR #1210, opened 2026-05-08
docs(rest): update empty load result data type to null

vmphase:
Empty object (`{ }`) => `null`, as returned from the Lavalink itself.


docs/api/rest.md
topi314: missing the `?` here

vmphase: Resolved.

topi314 (approved):

PR #1223, opened 2026-08-15
fix

Nansess:
discord voice connections have been shitter in the last couple of weeks resulting in lavalink locking up over time cause the voice connection joins forever so one xnio request gets fried then clients retry then more get stuck till all 640 are cooked and lavalink bricks


topi314:
prs go to dev, not master
rebaseee

Issue #1215, opened 2026-07-02
Cannot start Lavalink without IPv4


DRSchlaubi:
## Description
Lavalink tries to pull plugins from `maven.lavalink.dev` which is IPv4 only

## To Reproduce
1. Start Lavalink on an IPv6 only server

## Expected behavior
It works fine

Lavalink Logs:
```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'pluginManager': Failed to instantiate [lavalink.server.bootstrap.PluginManager$$SpringCGLIB$$0]: Constructor threw exception
	at org.springframework.beans.factory.support.ConstructorResolver.instantiate(ConstructorResolver.java:318) ~[spring-beans-6.1.15.jar!/:6.1.15]
	at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:306) ~[spring-beans-6.1.15.jar!/:6.1.15]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireConstructor(AbstractAutowireCapableBeanFactory.java:1375) ~[spring-beans-6.1.15.jar!/:6.1.15]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBeanInstance(AbstractAutowireCapableBeanFactory.java:1212) ~[spring-beans-6.1.15.jar!/:6.1.15]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:562) ~[spring-beans-6.1.15.jar!/:6.1.15]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:522) ~[spring-beans-6.1.15.jar!/:6.1.15]
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:337) ~[spring-beans-6.1.15.jar!/:6.1.15]
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:234) ~[spring-beans-6.1.15.jar!/:6.1.15]
```

## Additional context
Whilst the best way to fix it would be to add IPv6 support to maven.lavalink.dev, using NAT64 can also help, however by default the plugin downloader tries to use IPv4

PR #1229, opened 2026-09-07
Fix #1215: Configure java to use system default ip stack

Itz-Npg:
> Task :Lavalink-Server:test
lavalink.server.config.RequestAuthorizationFilterTest

  Test authenticatedRequest_Success() PASSED (1s)
  Test wrongAuthenticatedRequest_Fail() PASSED
  Test unauthenticatedRequest_Fail() PASSED

SUCCESS: Executed 3 tests in 8.8s
BUILD SUCCESSFUL in 29s


topi314:
That property is only read on startup 
So this has 0 impact

